PR #1, opened 2019-06-05
Added a new angular project for the digital library.  Added Chromium …

bgsavage:
…headless browser for running unit test.  Added a DataService and MockDataService which replaces it when the app is built with the standalone configuration.

Most of this code is scaffolded via `ng new`, but I'll call out parts that I've configured and added.


webapp/src/app/data/data.module.ts
```diff
@@ -0,0 +1,15 @@
...
+import { DataService } from 'src/app/data/data.service';
+import { HttpClientModule } from '@angular/common/http';
+
+@NgModule({
+  declarations: [ ],
+  imports: [
+    CommonModule, 
+    HttpClientModule
+  ],
+  providers: [{ provide: DataService, useClass: environment.dataService }]
```
bgsavage: Added this DataModule.  Here the DataService provider is defined by the environment.

webapp/src/app/data/data.service.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Wrapper service for Angular's HttpClient.  As a best practice, we shouldn't ever use
+ * HttpClient directly except for here, so that any future Angular HttpClient API changes
+ * can be updated to this one spot.  This also allows us the opportunity add any generic,
+ * Sbdl-specific logic in a single spot that should be applied across the board.
+ */
+@Injectable()
+export class DataService {
+
+  // TODO: Set and use CDL environment variable.
+  constructor(private httpService: HttpClient) { }
```
bgsavage: I never like to call `HttpClient` directly in my service classes for a few reasons.  This is just a generic DataService class I always include which wraps `HttpClient`.  For this project, I imagine we will handle environment configurations to call the external CDL api url, and provide a different different implementation of this class when running locally in a standalone mode.

webapp/src/environments/environment.ts
```diff
@@ -0,0 +1,20 @@
+import { DataService } from 'src/app/data/data.service';
+
+// This file can be replaced during build by using the `fileReplacements` array.
+// `ng build --prod` replaces `environment.ts` with `environment.prod.ts`.
+// The list of file replacements can be found in `angular.json`.
+
+export const environment = {
+  production: false,
+  standalone: false,
+  dataService: DataService
```
bgsavage: For environment.prod and the default environment, use the live DataService will will make external api calls.

webapp/src/environments/environment.standalone.ts
```diff
@@ -0,0 +1,8 @@
+import { MockDataService } from 'src/app/data/mock-data.service';
+
+export const environment = {
+    production: false,
+    standalone: true,
+    dataService: MockDataService
```
bgsavage: For standalone environment, use the MockDataService instead.

webapp/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "sb-digital-library",
+  "version": "0.0.0",
+  "scripts": {
+    "ng": "ng",
+    "start": "npm run test && ng serve",
+    "start-standalone": "npm run test && ng serve --c standalone",
+    "build": "npm run test && ng build --prod --aot --vendor-chunk",
+    "build-standalone": "npm run test && ng build --c standalone --aot --vendor-chunk",
```
bgsavage: Custom start and build commands will run the unit tests by default.

webapp/angular.json
```diff
@@ -0,0 +1,134 @@
...
+    "sb-digital-library": {
+      "projectType": "application",
+      "schematics": {
+        "@schematics/angular:component": {
+          "style": "scss"
+        }
+      },
+      "root": "",
+      "sourceRoot": "src",
+      "prefix": "sbdl",
```
bgsavage: Defined the app prefix to be `sbdl`. Thoughts?

webapp/angular.json
```diff
@@ -0,0 +1,134 @@
...
+              ]
+            },
+            "standalone" : {
+              "fileReplacements": [
+                {
+                  "replace": "src/environments/environment.ts",
+                  "with": "src/environments/environment.standalone.ts"
+                }
+              ]
+            }
```
bgsavage: Standalone configuration

webapp/angular.json
jdbernard: Works for me. 👍

webapp/src/app/data/data.service.ts
jdbernard: I tend to do the same. I agree with the approach of swapping in different implementations of the `DataService` class depending on which mode we're in.

PR #2, opened 2019-06-07
Added custom launchers for firefox and chrome, and debug-chrome and d…

bgsavage:
…ebug-firefox npm run commands to simplify unit test debugging.  Added -o flag on the start commands.

Note: pulling this change will require a fresh `npm install` to download the karma-firefox-launcher dependency.

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -3,11 +3,13 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "npm run test && ng serve",
-    "start-standalone": "npm run test && ng serve --c standalone",
+    "start": "npm run test && ng serve -o",
+    "start-standalone": "npm run test && ng serve --c standalone -o",
     "build": "npm run test && ng build --prod --aot --vendor-chunk",
     "build-standalone": "npm run test && ng build --c standalone --aot --vendor-chunk",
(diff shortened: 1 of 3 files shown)


webapp/package.json
```diff
@@ -3,11 +3,13 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "npm run test && ng serve",
-    "start-standalone": "npm run test && ng serve --c standalone",
+    "start": "npm run test && ng serve -o",
+    "start-standalone": "npm run test && ng serve --c standalone -o",
```
bgsavage: -o will automatically open the browser for us on startup.

webapp/package.json
```diff
@@ -3,11 +3,13 @@
...
     "build": "npm run test && ng build --prod --aot --vendor-chunk",
     "build-standalone": "npm run test && ng build --c standalone --aot --vendor-chunk",
     "test": "ng test --source-map=false --watch=false",
+    "test-debug-chrome": "ng test --browsers=ChromeDebug",
+    "test-debug-firefox": "ng test --browsers=FirefoxDebug",
```
bgsavage: So now instead of having go into the karma.conf.js and change the browsers array, I added custom launchers with npm run commands to call instead.  Once this gets merged I'll update the docs.

PR #3, opened 2019-06-10
Feature/setup configs

bgsavage:
Added an AppConfig service which loads all the settings found in a config.json file. Now loading AppConfig settings during app intiialization. Added a dev environment and including config.prod.json and config.dev.json to replace config.json when building depending on the build configuration.

Added unit tests

### DIFF
--- a/webapp/src/app/app.module.ts
+++ b/webapp/src/app/app.module.ts
@@ -1,9 +1,13 @@
+import { APP_INITIALIZER, NgModule } from '@angular/core';
 import { BrowserModule } from '@angular/platform-browser';
-import { NgModule } from '@angular/core';
-
 import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
+import { AppConfig } from './config/app.config';
 import { DataModule } from './data/data.module';
+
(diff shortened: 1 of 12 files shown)


webapp/src/app/app.module.ts
```diff
@@ -14,7 +18,11 @@ import { DataModule } from './data/data.module';
     AppRoutingModule,
     DataModule
   ],
-  providers: [],
-  bootstrap: [AppComponent]
+  providers: [{ 
+    provide: APP_INITIALIZER,
+    useFactory: initializeApp,
+    deps: [ AppConfig ], multi: true }
+  ],
```
bgsavage: Using the APP_INITIALIZER to ensure app config settings are loaded in at startup.  This allows us to reference configurations in the app via a static class `AppConfig.settings`

webapp/src/assets/config/config.prod.json
```diff
@@ -0,0 +1,12 @@
+{
```
bgsavage: We can define more later as needed, but for now, just created a default config, dev, and prod.

PR #4, opened 2019-06-14
Feature/app container

bgsavage:
This PR sets up some placeholders for the home page, resource page, and the header section.  Routing is included for each of the pages.  For the header, I used an example of an older instance of the sbac-ui-kit I had on my local thinking it might be a good starting point.  I temporarily pulled in and referenced that local sbac-ui-kit until I know more about how/if we will consuming the sbac-ui-kit.


webapp/src/app/app.component.spec.ts
```diff
@@ -30,10 +30,10 @@ describe('AppComponent', () => {
...
-    fixture.detectChanges();
-    const compiled = fixture.debugElement.nativeElement;
-    expect(compiled.querySelector('h1').textContent).toContain('Welcome to sb-digital-library!');
-  });
+  // it('should render title in a h1 tag', () => {
+  //   const fixture = TestBed.createComponent(AppComponent);
+  //   fixture.detectChanges();
+  //   const compiled = fixture.debugElement.nativeElement;
+  //   expect(compiled.querySelector('h1').textContent).toContain('Welcome to sb-digital-library!');
+  // });
```
bgsavage: Normally I'm not a proponent of checking in commented code, but wanted to leave the example of how we can test the app component once we know more about what the initial app component will be.

webapp/src/app/layout/app-container/app-container.component.html
```diff
@@ -0,0 +1,2 @@
+<sbdl-header></sbdl-header>
+<router-outlet></router-outlet>
```
bgsavage: I imagine this will also have some footer section.

webapp/src/app/app-routing.module.ts
```diff
@@ -1,10 +1,29 @@
...
 import { Routes, RouterModule } from '@angular/router';
+import { AppContainerComponent } from './layout/app-container/app-container.component';
+import { HomeComponent } from './home/home.component';
+import { ResourceComponent } from './resource/resource.component';
 
-const routes: Routes = [];
+const routes: Routes = [
+  {
+    path: '',
+    component: AppContainerComponent,
```
bgsavage: All routes using the app container (i.e. has the header, the future footer, maybe breadcrumbs, etc.) should be a child route below.  If it's not using the App container, it could be a sibling (for example, a help page or terms of use that may pop up in a new tab)

PR #6, opened 2019-06-18
Feature/add google analytics

bgsavage:


webapp/src/index.html
```diff
@@ -9,5 +9,13 @@
 </head>
 <body>
   <sbdl-root></sbdl-root>
+  <script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+    
+    ga('create', 'UA-142327690-1', 'auto');
```
bgsavage: I haven't found an easy way to make the analytics id configurable yet, so for now, just made the tracking itself configurablee.  The ID here is just an sbdl.dev@gmail.com account I created to test locally, but I know they will want to manage it themselves.  Once we have that ID, we can just replace it right here.

webapp/src/assets/config/config.prod.json
```diff
@@ -8,5 +8,6 @@
     },
     "apiServer": {
         "cdl": "https://prod.cdl.sbac.org/api/v1.0/"
-    }
+    },
+    "enableAnalytics": true
```
bgsavage: Only set to true for prod.

PR #7, opened 2019-06-20
Feature/site header navigation

bgsavage:
This PR adds a site header / navigation placeholder, sets up the initial infrastructure for calling and resolving user info, and adds unit tests.

Unrelated, removed the sbac-ui-kit that had been added previously for a previous placeholder for site header / navigation

image


webapp/src/app/data/mock-data.service.ts
```diff
@@ -1,23 +1,36 @@
...
-// wWrk around for: 
+// Work around for: 
 // https://stackoverflow.com/questions/48953587/typescript-class-implements-class-with-private-functions
 // tldr; We can't implement a class with privates, so we need a wrapper type to pull out the publics.
 type PublicPart<T> = {[K in keyof T]: T[K]}
 
 @Injectable()
 export class MockDataService implements PublicPart<DataService> {
+  readonly mockDataEndpoints = [
+    { pattern: /^\/userinfo$/, result: mockUser }
```
bgsavage: The idea is to use regex pattern matching to mock endpoints and return the matching data.

webapp/src/app/data/mock-data.service.ts
```diff
@@ -1,23 +1,36 @@
...
+  ];
+
   constructor() {
-    console.log('mock data service used.');
+    console.log('Mock data service loaded.');
```
bgsavage: Next thing I'd like to do is inject a logger rather than calling console.log directly.  I thought I saw somewhere we'd like to have logging to aws?

webapp/src/app/data/user/user.service.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { TestBed } from '@angular/core/testing';
+
+import { UserService } from './user.service';
+import { DataService } from '../data.service';
+import { MockDataService } from '../mock-data.service';
+
+describe('UserService', () => {
+  beforeEach(() => TestBed.configureTestingModule({
+    providers: [ { provide: DataService, useClass: MockDataService }]
```
bgsavage: Using the MockDataService for unit tests so I can assert on the expected below.

webapp/src/app/data/user/user.service.ts
```diff
@@ -0,0 +1,27 @@
...
+    return this.dataService
+      .get('/userinfo')
+      .pipe(map(x => this.mapToModel(x)));
+  }
+
+  private mapToModel(apiResult: any): UserModel {
+    return <UserModel> {
+      firstName: apiResult.firstName,
+      lastName: apiResult.lastName,
+      tenantName: apiResult.tenantName
```
bgsavage: Ideally once we do have a working API that returns actual user info, this should be the only part of the app that will get affected, where we map the API model to the UI model.

webapp/src/app/layout/header/header.component.html
```diff
@@ -1,34 +1,23 @@
...
-            <a href="#" class="dropdown-toggle btn-sm" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false"><i class="fa fa-user-circle"></i> Mary Anderson <span class="caret"></span></a>
-            <ul class="dropdown-menu">
-              <li><a href="#">Action</a></li>
-              <li><a href="#">Another action</a></li>
-              <li><a href="#">Something else here</a></li>
-              <li role="separator" class="divider"></li>
-              <li><a href="#">Logout</a></li>
-            </ul>
-          </li>
-        </ul>
```
bgsavage: This code above was just copy pasta from an older sbac-ui-kit I quickly threw up just to have something show up on screen.

webapp/src/app/layout/header/header.component.scss
```diff
@@ -0,0 +1,27 @@
...
+    padding-bottom: 25px;
+    width: 100%;
+
+    a {
+        color: #338225;
+        text-decoration: initial;
+        font-size: 1.2rem;
+        font-weight: bold;
+    }
+}
```
bgsavage: Placeholder styles that I didn't put a lot of thought in as it will be changed.  It's not even responsive right now.

webapp/src/app/layout/header/header.component.ts
```diff
@@ -7,9 +9,12 @@ import { Component, OnInit } from '@angular/core';
 })
 export class HeaderComponent implements OnInit {
 
-  constructor() { }
+  user: UserModel;
+
+  constructor(private route: ActivatedRoute) { }
 
   ngOnInit() {
+    this.user = this.route.snapshot.data.currentUser;
```
bgsavage: This is my internal debate right now.  Should the HeaderComponent be reaching to the route's resolve data to get the user info?  Or should the AppContainerComponent be responsible for reaching to the route data and passing down to the HeaderComponent? 

I went this way for now because it'll be easier to adapt later.

webapp/src/app/app-routing.module.ts
```diff
@@ -3,11 +3,15 @@ import { Routes, RouterModule } from '@angular/router';
...
 import { ResourceComponent } from './resource/resource.component';
+import { CurrentUserResolve } from './layout/current-user.resolve';
 
 const routes: Routes = [
   {
     path: '',
     component: AppContainerComponent,
+    resolve: {
+      currentUser: CurrentUserResolve
+    },
```
bgsavage: Resolve the current user when the root route is hit.

PR #8, opened 2019-06-21
Feature/logging

bgsavage:
Added a logging service so later logs can be shipped outside the angular app.  Added a global error handler which consumes the logging service.


webapp/src/app/common/logging/logging.service.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  log(severity: LoggerLevel, message: string, params?: any) {
+    if(severity >= this.configuredLevel) {
+      const consoleLogger = this.consoleSeverityLoggers.find(x => x.severity === severity);
+      if(consoleLogger) {
+        if(params !== null && params !== undefined) {
+          consoleLogger.logParams(message, params);
+        } else {
+          consoleLogger.log(message);
+        }
```
bgsavage: TODO: Add graylog api call here.

PR #14, opened 2019-07-23
Feature/material components

bgsavage:
Now using material components for our buttons.  Added wrapper component to facilitate usages of buttons and button-icons.

image


webapp/angular.json
```diff
@@ -35,7 +35,8 @@
             ],
             "stylePreprocessorOptions": {
               "includePaths": [
-                "src/ui-kit"
+                "src/ui-kit",
+                "node_modules"
```
bgsavage: This allows us to @import scss modules from the node_modules directory.  Required for bringing in the material styles.

PR #16, opened 2019-07-25
Implement Demo Feedback

bgsavage:
Now displaying each claim, target, standard on its own line in the case of multiples.  Now this text will have ellipsis when overflowing.

image

### DIFF
--- a/webapp/src/app/resource/metadata/metadata.component.html
+++ b/webapp/src/app/resource/metadata/metadata.component.html
@@ -1,4 +1,4 @@
-<div class="metadata-component">
+<div class="metadata-container">
   <div class="image-thumbnail">
     <img [src]="model.image">
   </div>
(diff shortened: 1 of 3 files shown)


webapp/src/app/resource/metadata/metadata.component.html
```diff
@@ -11,21 +11,9 @@ <h6 class="grades">{{ model.grades | join }}</h6>
...
-    </label>
-    <div class="value">{{ model.targets | join:{ field: 'title' } }}</div>
-    <label [ngPlural]="model.standards.length">
-      <ng-template ngPluralCase="=1">Standard</ng-template>
-      <ng-template ngPluralCase="other">Standards</ng-template>
-    </label>
-    <div class="value">{{ model.standards | join }}</div>
+    <ng-container *ngTemplateOutlet="alignment; context: { models: model.claims, title: 'Claim', titles: 'Claims' }"></ng-container>
+    <ng-container *ngTemplateOutlet="alignment; context: { models: model.targets, title: 'Target', titles: 'Targets' }"></ng-container>
+    <ng-container *ngTemplateOutlet="alignment; context: { models: model.standards, title: 'Standard', titles: 'Standards' }"></ng-container>
```
bgsavage: Moved to an ng-template since this code is similar each of these 3 fields and is becoming quite verbose.

PR #17, opened 2019-07-26
Feature/toggle favorite

bgsavage:
Added favorite toggling of resource with a simulated backend.

### DIFF
--- a/webapp/.vscode/settings.json
+++ b/webapp/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.exclude": {
+        "node_modules": true
+    }
+}
--- a/webapp/src/app/common/controls/button-icon/button-icon.component.html
+++ b/webapp/src/app/common/controls/button-icon/button-icon.component.html
@@ -1,3 +1,3 @@
-<button #button class="mdc-icon-button material-icons">
+<button #button class="mdc-icon-button material-icons" [disabled]="disabled">
   <ng-content></ng-content>
 </button>
--- a/webapp/src/app/data/favorite/favorite.service.ts
+++ b/webapp/src/app/data/favorite/favorite.service.ts
@@ -1,0 +1,34 @@
+import { Injectable } from '@angular/core';
+import { DataService } from '../data.service';
+import { FavoriteResource } from './model/favorite-resource.model';
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class FavoriteService {
+
+  constructor(private dataService: DataService) { }
+
+  postFavoriteResource(favoriteResource: FavoriteResource): Observable<FavoriteResource> {
+    const apiModel = this.mapFavoriteResourceToApi(favoriteResource);
+    return this.dataService
+      .post('/favorites/resource', apiModel)
+      .pipe(map(res => this.mapApiToFavoriteResource(res)));
+  }
+
+  private mapFavoriteResourceToApi(uiModel: FavoriteResource) {
+    return {
+      resourceId: uiModel.resourceId,
+      favorite: uiModel.favorite
+    };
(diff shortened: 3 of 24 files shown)


webapp/.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "files.exclude": {
+        "node_modules": true
```
bgsavage: This is just a vs code setting to not add node_modules to the file tree.  I got annoyed by always getting sucked into the node_modules vacuum.  You can ignore if you're not using vs code.

webapp/src/app/common/controls/button-icon/button-icon.component.html
```diff
@@ -1,3 +1,3 @@
-<button #button class="mdc-icon-button material-icons">
+<button #button class="mdc-icon-button material-icons" [disabled]="disabled">
```
bgsavage: I was going to disable the fav button on click while it was talking to the API, but ended up replacing with a loading indicator.  I didn't back out the disabled code in case we need it later?

webapp/src/app/data/favorite/favorite.service.ts
```diff
@@ -0,0 +1,34 @@
...
+      resourceId: uiModel.resourceId,
+      favorite: uiModel.favorite
+    };
+  }
+
+  private mapApiToFavoriteResource(apiModel: any): FavoriteResource {
+    return {
+      resourceId: apiModel.resourceId,
+      favorite: apiModel.favorite
+    }
```
bgsavage: Yes api and ui models are identical right now, but have no idea what the api will look like so this is just to provide a layer of protection so when we do get the api model, should only have to update the mappers here.

webapp/src/app/data/mock-data.service.ts
```diff
@@ -11,30 +12,61 @@ type PublicPart<T> = {[K in keyof T]: T[K]}
...
+
   downloadPdf(url: string): Observable<Blob> {
     throw new Error("Method not implemented.");
   }
+
+  private setFavorite(object: any): any{
+    
+    const resource = this.resources.find(x => x.id === object.resourceId);
+    if(resource) {
+      resource.favorite = object.favorite;
```
bgsavage: Setting the favorite here simulates persistence, but in memory only.  As soon as you refresh the page it's reset back to original.

webapp/src/app/data/mock-data.service.ts
```diff
@@ -11,30 +12,61 @@ type PublicPart<T> = {[K in keyof T]: T[K]}
...
 
     return of(undefined);
   }
 
+  post(url: string, obj: any): Observable<any> {
+    const mockedEndpoint = this.mockPostDataEndpoints.find(x => url.match(x.pattern) && url.match(x.pattern).length !=0 );
+
+    if(mockedEndpoint && mockedEndpoint.post) {
+      this.logger.debug(`Mocking API POST for ${url}`, obj);
+      return of(mockedEndpoint.post(obj)).pipe(delay(500));
```
bgsavage: Half second delay is more realistic than instant.

webapp/src/app/layout/app-container/app-container.component.html
```diff
@@ -2,6 +2,9 @@
     <div class="navigation">
         <!-- TODO: Add navigation -->
         (nav)
+        <p><sbdl-button class ="primary" [routerLink]="['']">Home</sbdl-button></p>
+        <p><sbdl-button class ="primary" [routerLink]="['/resource/1']">One</sbdl-button></p>
+        <p><sbdl-button class ="primary" [routerLink]="['/resource/2']">Two</sbdl-button></p>
```
bgsavage: Added these buttons as temp nav elements to allow me to navigate between resources without invoking a full app reload (and losing my in-memory persistence of a favorited resource)

PR #20, opened 2019-07-31
Feature/share action button

bgsavage:
This PR implements the share action functionality

image

### DIFF
--- a/webapp/src/app/common/controls/popover/popover.component.html
+++ b/webapp/src/app/common/controls/popover/popover.component.html
@@ -1,0 +1,7 @@
+<div class="popover-parent">
+    <div role="tooltip" class="popover-container">
+        <div class="popover-content">
+          <ng-container *ngTemplateOutlet="template"></ng-container>
+        </div>
+      </div>
+</div>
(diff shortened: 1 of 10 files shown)


webapp/src/app/common/controls/popover/popover.component.html
```diff
@@ -0,0 +1,7 @@
+<div class="popover-parent">
+    <div role="tooltip" class="popover-container">
+        <div class="popover-content">
+          <ng-container *ngTemplateOutlet="template"></ng-container>
+        </div>
+      </div>
+</div>
```
bgsavage: Rolling our own popover component since it's not available in material web components yet.

webapp/src/app/common/controls/popover/popover.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  ngOnInit(): void {
+  }
+
+  private parentNode: any;
+
+  @Output()
+  onClose = new EventEmitter();
+
+  @Input()
+  template;
```
bgsavage: Our popover at the moment requires you to pass the template.  Assumption here is if you don't have a template to pass, you should be using a tooltip instead.

PR #25, opened 2019-08-06
Created a read-more component which truncates text if it's longer tha…

bgsavage:
…n the defined const.  Applied this new component to the overview section so that the description is truncated.

Screen Shot 2019-08-06 at 9 52 52 AM

Screen Shot 2019-08-06 at 9 53 02 AM

### DIFF
--- a/webapp/src/app/common/controls/read-more/read-more.component.ts
+++ b/webapp/src/app/common/controls/read-more/read-more.component.ts
@@ -1,0 +1,35 @@
+import { Component, OnInit, Input, OnChanges } from '@angular/core';
+
+@Component({
+  selector: 'sbdl-read-more',
+  templateUrl: './read-more.component.html'
+})
+export class ReadMoreComponent implements OnInit, OnChanges {
+  @Input()
+  text: string;
+
+  readonly MaxNumberOfCharacters = 168;
+  displayText: string;
+
+  collapsed: boolean = true;
+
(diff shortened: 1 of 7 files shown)


webapp/src/app/common/controls/read-more/read-more.component.ts
```diff
@@ -0,0 +1,35 @@
...
+
+@Component({
+  selector: 'sbdl-read-more',
+  templateUrl: './read-more.component.html'
+})
+export class ReadMoreComponent implements OnInit, OnChanges {
+  @Input()
+  text: string;
+
+  readonly MaxNumberOfCharacters = 168;
```
bgsavage: Not sure what this actually should be, so might as well match the mockup for now.  We can discuss making this configurable later, or if we need, it'd be easy enough to just change the constant here.

webapp/src/ui-kit/_typography.scss
```diff
@@ -102,3 +104,8 @@ p, ol, ul, table {
 .emphasis-medium {
     opacity: 0.65;
 }
+
+a {
+    color: $action-700;
+    text-decoration: none;
```
bgsavage: @aaronsleeper  The more / less is the only normal "anchor" I've seen in the app, so I assume this is global.  Let me know if I need to scope it just to the read-more component.

PR #29, opened 2019-08-08
Move CI/CD to CodeBuild only, no longer using CodePipeline

jdbernard:
There are some limitations in CodePipeline that make it unsuitable for
our workflow. In particular:

1. There is no way to create a pipeline that is only triggered manually.
   This is something we will need in order to automate our deployments
   to environments like QA and UAT (really anything other than dev).

2. There is no way to dynamically set the source revision that
   CodePipeline pulls. When you define a Source set in CP you have to
   give a branch name. Again, this makes it impossible to specify a
   specific revision or tagged version.

3. When CodePipeline creates the Source artifact bundle that gets passed
   to CodeBuild it excludes the .git directory. This means any commands
   in our build process that rely on the repo being present will fail.
   For example version stamping calls `git describe`.

4. CodePipeline does not support PR builds. It naively looks for any new
   commits on the configured branch. So in addition to being unable to
   trigger a pipeline with a specific revision, CP cannot react to
   webhooks for automated PR builds.

In order to support the workflows we want we are relying solely on
CodeBuild and are building the deploy process into the build process.

* The buildspec.yaml definition now includes a `post_build` phase which
  calls our new deploy.sh script.
* The deploy script runs after the build has completed successfully and
  inspects the build environment to decide between three scenarios:

  - PR merged to dev -> deploy to dev environment.
  - PR created/updated/reopened -> no deploy
  - Manually triggered build and deploy -> deploy to the environment
    named in the manual call (specified using the `DL_TARGET_ENV`
    environment variable).

  After identifying the target environment (if any) and performing the
  deploy (using the AWS CLI's S3 sync command), the deploy script
  queries the CloudFront API to lookup the distribution ID for that
  environment (using the environment domain name) and creates an
  invalidation for the /index.html asset.

Because the CodeBuild project is not managed in our Terraform
configuration, this will require a manual adjustment to the CodeBuild
project to enable GitHub webhook integrations.


webapp/package.json
```diff
@@ -6,7 +6,7 @@
     "start": "npm run pre-build && ng serve -o",
     "start-standalone": "npm run pre-build && ng serve --c standalone -o",
     "build": "npm run pre-build && ng build --prod --aot --vendor-chunk",
-    "build-ci": "npm run pre-build-ci && ng build --prod --aot --vendor-chunk",
+    "build-ci": "npm run pre-build-ci && ng build --c standalone --prod --aot --vendor-chunk",
```
jdbernard: We may want to talk about how we manage this going forward. Ideally I'd like to not have to make a code-change to our build scripts to switch between hitting our mock API endpoints vs. the life API. I don't have an immediate solution in mind, it's something I'd like to think about in the future.

bgsavage: I believe you should be able to pass custom params from the build something like `npm run build-ci [--c standalone]` ( I haven't tried myself).  That being said, I think the `npm run build` should be the build-ci.  Developer should be able to run `npm run build` with confidence that if it passes locally, it should also pass on the CI server.  And if it fails on the CI server, it should fail for the same reason locally (facilitates debugging build breaks).

We can make the --c standalone part of the default build, my concern about putting it in there initially was forgetting about it later when we actually have a live API.  But that is the default build right now so we should put it, just need to be careful about removing later when abandon the mock data service.

jdbernard: The reason for the separation between `build` and `build-ci` is that on the CI server we need to run Headless Chrome without sandboxing, whereas we probably want it on our local machines because we're not getting the built-in sandboxing that Docker provides on the CI server.

I think no matter where we put the `--c standalone` it's going to result in a code change or config change to update. I'll test it as a param to `npm run`. I think I like it better there rather than in the config file.

jdbernard: Looks like we need to rethink that. Moving it out of the script and into the invocation caused the build to fail. We'll revisit after the current sprint.

PR #30, opened 2019-08-08
Terraform Environment Configuration

jdbernard:
Terraform to standup and manage all of our operational environments (with the exception of PROD which is TBD). :rocket:


operations/terraform/terraform.tfstate
```diff
@@ -0,0 +1,1323 @@
+{
```
jdbernard: .tfstate files store Terraform's understanding of the current state of the deployed infrastructure. Anyone who will be applying Terraform configuration will need the up-to-date version of this file. The repo is not the best place for this to live, but here it will be until we have time to implement and test remote (S3) storage of the file.

See [DL-65 Move Terraform State Storage to a Remote Backend](https://smarterbalanced.atlassian.net/browse/DL-65)

PR #31, opened 2019-08-12
Feature/demo feedback

bgsavage:
This PR addresses various minor aesthetics discussed during the demo.
  - Additional padding when an element is scrolled to
  - Added read-more control to learning goal should it go over
  - Fixed an issue with tabbing the outline when in reading mode would throw off margins /padding
  - Attachment card is now clickable and has the mdc ripple
  - Attachments button at in the content section now works

### DIFF
--- a/webapp/src/app/common/controls/read-more/read-more.component.spec.ts
+++ b/webapp/src/app/common/controls/read-more/read-more.component.spec.ts
@@ -1,7 +1,5 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
-
 import { ReadMoreComponent } from './read-more.component';
-import { By } from '@angular/platform-browser';
 
 describe('ReadMoreComponent', () => {
   let component: ReadMoreComponent;
--- a/webapp/src/app/common/controls/read-more/read-more.component.spec.ts
+++ b/webapp/src/app/common/controls/read-more/read-more.component.spec.ts
@@ -1,6 +1,7 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { ReadMoreComponent } from './read-more.component';
+import { By } from '@angular/platform-browser';
 
 describe('ReadMoreComponent', () => {
   let component: ReadMoreComponent;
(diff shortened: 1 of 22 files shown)


webapp/src/app/common/controls/read-more/read-more.component.spec.ts
```diff
@@ -22,4 +21,46 @@ describe('ReadMoreComponent', () => {
...
+
+    component.toggleCollapsed();
+    fixture.detectChanges();
+
+    const actual = fixture.debugElement.nativeElement
+    expect(actual.textContent).toContain(component.text);
+    expect(actual.querySelector('a').textContent).toContain('less');        
+  });
+
+  it('should hide more / less link when given text is less than MaxNumberOfCharacters', ()=> {
```
bgsavage: Added unit tests

webapp/src/app/common/controls/read-more/read-more.component.ts
```diff
@@ -1,35 +1,53 @@
...
-  @Input()
-  truncated: boolean;
+  private _text: string;
 
   constructor() { }
 
   ngOnInit() {
   }
 
-  ngOnChanges() {
```
bgsavage: Using ngOnChanges actually makes it more difficult to unit test if I want to test @Input changes.  So I changed to getter / setter instead.

webapp/src/app/resource/instructional/attachments/attachments.component.html
```diff
@@ -3,16 +3,14 @@ <h4 class="leading-icon">
       <i class="far fa-paperclip"></i>Attachments
   </h4>
   <p>This resource relies on the following materials. Download them to use.</p>
-  <div *ngFor="let attachment of models" class="attachment-card">
+  <a *ngFor="let attachment of models" [href]="attachment.downloadUrl" [download]="attachment.filename"  class="attachment-card"  #attachments>
```
bgsavage: Moved the anchor tag up to the entire block since the entire card should be clickable.

webapp/src/app/resource/instructional/content/resource-content.component.scss
```diff
@@ -6,20 +6,12 @@
...
-        height: 46px;
-        background-image: url('./../../../../assets/images/yellow-spot.png');
-        background-position: top left;
-        background-size: cover;
-        background-repeat: no-repeat;
+    .top-header {
+        display: flex;
+        justify-content: space-between;
+        width: 100%;
+        align-items: baseline;
```
bgsavage: I went back to using a `<img src` block because the `:before` and `position: absolute` would move the yellow-spot over the outline when resizing the window.

webapp/src/ui-kit/_buttons.scss
```diff
@@ -61,11 +61,12 @@
...
     border-left: $action-500 solid 4px;
+    margin-left: -19px;
 }
 
 .step > .mdc-button:before,
 .step > .mdc-button:after,
 .outline >.mdc-button:before,
 .outline > .mdc-button:after {
-  background-color: $action-300;
+  background-color: $action-500;
```
bgsavage: Bumped the outline hover color to make it more visible.

PR #34, opened 2019-08-16
Feature/integrate api

bgsavage:
This PR creates a "stage" build / run mode which allows for the DL 2.0 front end to directly pull data from the CDL stage environment.  Note: There is a manual step to generate your own auth token and place in the config.stage.json file as the token has an expiration of 24hrs.


webapp/angular.json
```diff
@@ -74,6 +74,33 @@
                 }
               ]
             },
+            "stage": {
+              "fileReplacements": [
+                {
+                  "replace": "src/environments/environment.ts",
+                  "with": "src/environments/environment.prod.ts"
```
bgsavage: Stage build uses prod environment as it should be "prod"-like.   When you run `npm run start-stage`, you'll get a warning to not use this in actual production.  This is expected.

webapp/src/app/app-routing.module.ts
```diff
@@ -1,16 +1,15 @@
...
+import { AppContainerComponent } from './layout/app-container/app-container.component';
 import { ResourceTypeStrategyComponent } from './resource/resource-type-strategy.component';
+import { ResourceResolve } from './resource/resource.resolve';
 
 const routes: Routes = [
   {
     path: '',
     component: AppContainerComponent,
-    resolve: { currentUser: CurrentUserResolve },
+    // resolve: { currentUser: CurrentUserResolve },
```
bgsavage: I'm not sure what this will end up being, so commenting out for now as placeholder.

webapp/src/app/common/config/config.model.ts
```diff
@@ -3,6 +3,6 @@ import { LoggerLevel } from '../logging/logger-level.enum';
 export interface IAppConfig {
     env: { name: string },
     logging: { console: boolean, aws: boolean, level: LoggerLevel },
-    apiServer: { cdl: string },
+    apiServer: { cdl: string, authToken: string },
```
bgsavage: Not a permanent solution to have our auth token in a config file, but useful for now in dev.

webapp/src/app/data/data.service.ts
```diff
@@ -31,24 +27,44 @@ export class DataService {
...
+      headers: new HttpHeaders({
+        'Content-Type':  'application/octet-stream',
+        'Authorization': 'Bearer ' + AppConfig.settings.apiServer.authToken
+      }),
+      params: params,
+      responseType : 'arraybuffer',
+    };
+
+    return this.httpService.get(fullUrl, options)
+      .pipe(map(response => new Blob([ response ])))
```
bgsavage: map the response as a blob.

webapp/src/app/data/mock-data.service.ts
```diff
@@ -14,15 +15,23 @@ type PublicPart<T> = {[K in keyof T]: T[K]}
...
+    { pattern: /\/resources\/0$/, result: mockApiResourceWithNulls },
+    { pattern: /\/resources\/2$/, result: mockApiResource2 },
+    { pattern: /\/resources\/[0-9]/, result: mockApiResource },
+    { pattern: /\/file_documents\/52/, result: mockDocument52 },
+    { pattern: /\/file_documents\/53/, result: mockDocument53 }
   ];
 
   readonly mockPostDataEndpoints = [
-    { pattern: /^\/favorites\/resource$/, post: body => this.setFavorite(body) },
+    { pattern: /^\/favorites\/resources$/, post: body => this.setFavorite(body) },
```
bgsavage: Hm, don't think this should have changed.  Looking now.

bgsavage: (My 's' key is broken and often times i accidentally leave stray s's now when i'm trying to save a file)

webapp/src/app/data/mock-data.ts
```diff
@@ -75,8 +75,8 @@ export const mockApiResourceExample = {
 	licenseSecondary: "Creative Commons Attribution",
 	copyrightClearaceForms: "",
 	pii: "None",
-	subjects: ["ELA", "Math"],
-	grades: ["Grade 6", "Grade 8", "Grade 9"],
+	subject: ["ELA", "Math"],
+	grade: ["Grade 6", "Grade 8", "Grade 9"],
```
bgsavage: updates to the api model which were actually being returned from the api in uat.

webapp/src/app/data/mock-data.ts
```diff
@@ -93,6 +93,10 @@ export const mockApiResourceExample = {
 	supportingEvidence: "",
 	principles: "",
 	mediaType: ["Video"],
+	documents: [
+		"/api/file_documents/53/download",
+        "/api/file_documents/52/download"
+	],
```
bgsavage: @jdbernard  
These are the attachments returned from the api.  So in the resource service I have to extract out the Ids here in order to make subsequent calls to `api/file_documents/{id}` to get more info about the attachments.

bgsavage: (This is error prone if the format of this url changes, etc.)

webapp/src/app/data/mock-data.ts
```diff
@@ -562,4 +567,24 @@ const polyfillMissingApiData2 = {
...
+
+export const mockDocument53 = {
+	"@context": "/api/contexts/FileDocument",
+	"@id": "/api/file_documents/53",
+	"@type": "FileDocument",
+	"id": 53,
+	"name": "note_taking.docx",
+	"path": "15d4a90dba2441666f775be72bdbd1e39816d684.docx",
+	"mimeType": null
+}
```
bgsavage: Added mock documents as returned from the stage api.

webapp/src/app/data/resource/resource.service.ts
```diff
@@ -29,21 +31,49 @@ export class ResourceService {
...
   get(id: number): Observable<ResourceModel> {
+    let resourceModel: any;
     return this.dataService
-      .get(`/resource/${id}`)
-      .pipe(map(apiModel => this.mapToResourceModel(apiModel)));
+      .get(`/resources/${id}`)
+      .pipe(mergeMap(apiModel => {
+        resourceModel = apiModel;
+        return apiModel.documents && apiModel.documents.length > 0
+          ? forkJoin(this.getAttachments(apiModel.documents))
```
bgsavage: @jdbernard 
If there's anything in the documents array, here I have to make additional calls to each `api/file_documents/{id}` endpoint in that array.  This is slows down the initial page load significantly, and adds increased load to the API if we have to make X number of calls on each page load.

webapp/src/app/data/resource/resource.service.ts
```diff
@@ -29,21 +31,49 @@ export class ResourceService {
...
+      })); 
+  }
+
+  getAttachments(documents: string[]): Observable<any>[] {
+    let observables = [];
+    
+    for(let doc of documents) {
+      const regexGroups = doc.match(/\/file_documents\/([0-9]*)\/download/);
+      if(regexGroups && regexGroups.length > 1){
+        const id = +regexGroups[1];
```
bgsavage: @jdbernard 
Here's where I extract out the id from the `file_documents/{id}/download` string. and then delegate down to my attachment service to make the api call.

webapp/src/assets/config/config.stage.json
```diff
@@ -0,0 +1,15 @@
...
+        "name": "DEV"
+     },
+    "logging": {
+        "console": true,
+        "aws": false,
+        "level": 0
+    },
+    "apiServer": {
+        "cdl": "https://api-stage.dl.smarterbalanced.org/api",
+        "authToken": "--omit--"
```
bgsavage: Here's where you plug in the auth token you would manually generate by calling the oauth/v2/token endpoint and passing client id and secret.  Tokens expire after 24 hrs.

webapp/src/app/resource/instructional/attachments/attachments.component.ts
```diff
@@ -38,4 +40,21 @@ export class AttachmentsComponent implements OnInit {
...
+  download(attachment: AttachmentModel) {
+    this.service
+      .download(attachment.id)
+      .subscribe(blob => {
+        const link = document.createElement('a');
+        link.href = window.URL.createObjectURL(blob);
+        link.download = attachment.filename;
+        link.hidden = true;
+        document.body.appendChild(link);
+        link.click();
```
bgsavage: Unfortunately there's not a more elegant way to download blobs from an API without having to create and remove a dom anchor element.  I'm open to alternatives if you know of one though.

PR #35, opened 2019-08-19
Feature/update read more

bgsavage:
This PR updates the read-more component to act on vertical height rather than number of characters.


webapp/src/app/common/controls/read-more/read-more.component.scss
```diff
@@ -0,0 +1,9 @@
+.text-container {
+    display: block;
+    overflow-y: hidden;
+    --max-height: 44px;
+
+    &.collapsed {
+      max-height:  var(--max-height);
```
bgsavage: I was trying to add a transition so that the expand / collapse would have an animation.  However, since the expanded doesn't have a max-height, and we wouldn't want to guess one, it's not possible animate.  I did spend some time seeing if I could calculate the rendered height in ts on load and set the max-height then via a css var, however it started to become too complex especially if you have to consider screen resizing.

webapp/src/app/common/controls/read-more/read-more.component.ts
```diff
@@ -22,32 +25,38 @@ export class ReadMoreComponent implements OnInit {
...
     this.collapsed = !this.collapsed;
-    this.toggleDisplayText();
   }
 
   private onTextChanges() {
-    this.collapsible = this.text && this.text.length > this.MaxNumberOfCharacters;
-    this.toggleDisplayText();
+    if(this.textContainer && this.textContainer.nativeElement) {
+      const contentHeight = this.textContainer.nativeElement.offsetHeight;
+      this.collapsible = contentHeight > this.maxHeightInPx;
```
bgsavage: Getting the content height here is useful only for determining if it is collapsible.

PR #37, opened 2019-08-21
Feature/professional resource init

bgsavage:
This PR restructures components that were assumed to be specific to IR for more generic reuse and then creates an initial template of the professional resource which can reuse all of those shared components.


bgsavage:
There's a lot to review here with structural changes.  Let me know if you'd like to do a live walkthrough of the major changes.

PR #116, opened 2020-01-13
DL-271: Support for deep-linking into resource documents.

jdbernard:


webapp/src/app/app-routing.module.ts
```diff
@@ -58,7 +58,7 @@ const routes: Routes = [
 ];
 
 @NgModule({
-  imports: [ RouterModule.forRoot(routes) ],
+  imports: [ RouterModule.forRoot(routes, { anchorScrolling: 'enabled' }) ],
```
jdbernard: Theoretically enabling [anchorScrolling](https://angular.io/api/router/ExtraOptions#anchorScrolling) should cause the router to automatically scroll to anchor fragments (`#step-1` for example) that are present in routes. However in testing this does not seem to be working. My initial thoughts are that it either:
1. related to our resource component structure: maybe anchorScrolling doesn't apply to child-component navigation within the same parent component 
2. related to the dynamic nature of our resource components: maybe the IDs for the anchors don't exist at the time when the anchorScrolling functionality in the router is called.

PR #138, opened 2020-01-17
[DL-280] Updated font size

jmesches-fairway:


jmesches-fairway:
This issue was fixed through PR for DL-254

PR #165, opened 2020-01-28
fixes grade filter order

esotrope:
image

### DIFF
--- a/webapp/src/app/data/mock-data.ts
+++ b/webapp/src/app/data/mock-data.ts
@@ -113,8 +113,8 @@ export const initialSearchFilters: SearchFilters = {
     { code: 'ir', title: 'Instructional' },
     { code: 'cp', title: 'Connections Playlist' },
     { code: 'pl', title: 'Professional Learning' },
-    { code: 'as', title: 'Accessibility Strategy' },
-    { code: 'fs', title: 'Formative Assessment Strategy' }
+    { code: 'fs', title: 'Formative Assessment Strategy' },
+    { code: 'as', title: 'Accessibility Strategy' }
   ],
   standards: [
     { code: 'ela-g3-l3a',     title: 'L-3a' },
--- a/webapp/src/app/data/search/search.service.ts
+++ b/webapp/src/app/data/search/search.service.ts
@@ -1,27 +1,108 @@
-import { Injectable } from '@angular/core';
-import { Observable, of, from } from 'rxjs';
-import { expand } from 'rxjs/internal/operators/expand';
-import { map } from 'rxjs/internal/operators/map';
-import { mergeMap } from 'rxjs/internal/operators/mergeMap';
-import { reduce } from 'rxjs/internal/operators/reduce';
-import { takeWhile } from 'rxjs/internal/operators/takeWhile';
-import { tap } from 'rxjs/internal/operators/tap';
-import { EMPTY } from 'rxjs';
-import { DataService } from '../data.service';
-import { SearchResults } from './search-results.model';
-import { SearchRequestModel } from './search-request.model';
-import { coalesce } from 'src/app/common/utils';
-import { ResourceService } from '../resource/resource.service';
-import { Claim } from '../resource/model/claim.model';
-import { Grade } from '../resource/model/grade.model';
-import { Subject } from '../resource/model/subject.model';
-import { Standard } from '../resource/model/standard.model';
-import { Target } from '../resource/model/target.model';
-import { SearchFilters, Filter, emptyFilters } from './search-filters.model';
-import { ResourceType } from '../resource/model/resource-type.enum';
-import { ResourceSummary } from '../resource/model/summary.model';
-import { fastArrayMerge } from 'src/app/common/utils';
-import { initialSearchFilters } from '../mock-data';
+import {Injectable} from '@angular/core';
+import {EMPTY, forkJoin, Observable, of} from 'rxjs';
+import {expand} from 'rxjs/internal/operators/expand';
+import {map} from 'rxjs/internal/operators/map';
+import {reduce} from 'rxjs/internal/operators/reduce';
+import {DataService} from '../data.service';
+import {SearchResults} from './search-results.model';
+import {SearchRequestModel} from './search-request.model';
+import {fastArrayMerge} from 'src/app/common/utils';
+import {Claim} from '../resource/model/claim.model';
+import {Grade} from '../resource/model/grade.model';
+import {Subject} from '../resource/model/subject.model';
+import {Standard} from '../resource/model/standard.model';
+import {Target} from '../resource/model/target.model';
+import {Filter, SearchFilters} from './search-filters.model';
+import {ResourceSummary} from '../resource/model/summary.model';
+import {initialSearchFilters} from '../mock-data';
+
+interface BumpySearchResults {
+  results: ResourceSummary[][];
+  filters: SearchFilters[][];
+}
+
+/**
+ * String filter code comparator
+ */
+const byCode = (a: Filter, b: Filter) => a.code.localeCompare(b.code);
+
+
+/**
+ * Creates and returns a comparator given the rankings
+ * @param rankings The ordered set of primitives of the same type to use for sorting
+ */
+function byRanking<T>(rankings: T[]): (a: T, b: T) => number {
+  return (a, b) => {
+    const indexA = rankings.findIndex(x => x === a);
+    const indexB = rankings.findIndex(x => x === b);
+    if (indexA < indexB) {
+      return -1;
+    }
+    if (indexA > indexB) {
+      return 1;
+    }
+    return 0;
+  };
+}
+
+/**
+ * Ensures there is no duplication in values for the given array
+ * @param values The values to de-duplicate
+ */
+function toArraySet<T>(values: T[]): T[] {
+  return Array.from(new Set(values));
+}
+
+/**
+ * Returns a sorted collection of the search filters given the provided filters and their defaults
+ * @param filters The current search filters to sort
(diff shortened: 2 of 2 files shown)


webapp/src/app/data/mock-data.ts
```diff
@@ -113,8 +113,8 @@ export const initialSearchFilters: SearchFilters = {
     { code: 'ir', title: 'Instructional' },
     { code: 'cp', title: 'Connections Playlist' },
     { code: 'pl', title: 'Professional Learning' },
-    { code: 'as', title: 'Accessibility Strategy' },
-    { code: 'fs', title: 'Formative Assessment Strategy' }
+    { code: 'fs', title: 'Formative Assessment Strategy' },
+    { code: 'as', title: 'Accessibility Strategy' }
```
esotrope: will make sense later

webapp/src/app/data/search/search.service.ts
```diff
@@ -1,27 +1,108 @@
...
+import {Claim} from '../resource/model/claim.model';
+import {Grade} from '../resource/model/grade.model';
+import {Subject} from '../resource/model/subject.model';
+import {Standard} from '../resource/model/standard.model';
+import {Target} from '../resource/model/target.model';
+import {Filter, SearchFilters} from './search-filters.model';
+import {ResourceSummary} from '../resource/model/summary.model';
+import {initialSearchFilters} from '../mock-data';
+
+interface BumpySearchResults {
```
esotrope: moved to the top - convention is to have constants/interfaces declared before their use

webapp/src/app/data/search/search.service.ts
```diff
@@ -1,27 +1,108 @@
...
+ * String filter code comparator
+ */
+const byCode = (a: Filter, b: Filter) => a.code.localeCompare(b.code);
+
+
+/**
+ * Creates and returns a comparator given the rankings
+ * @param rankings The ordered set of primitives of the same type to use for sorting
+ */
+function byRanking<T>(rankings: T[]): (a: T, b: T) => number {
```
esotrope: util for creating a ranking style comparator

webapp/src/app/data/search/search.service.ts
```diff
@@ -1,27 +1,108 @@
...
+    }
+    return 0;
+  };
+}
+
+/**
+ * Ensures there is no duplication in values for the given array
+ * @param values The values to de-duplicate
+ */
+function toArraySet<T>(values: T[]): T[] {
```
esotrope: probably functionally duplicative with the original "dedup" method i renamed to "mergeSearchFilters"

webapp/src/app/data/search/search.service.ts
```diff
@@ -1,27 +1,108 @@
...
+  return Array.from(new Set(values));
+}
+
+/**
+ * Returns a sorted collection of the search filters given the provided filters and their defaults
+ * @param filters The current search filters to sort
+ * @param filterOptions The initial search filter options and their natural ordering
+ */
+function normalizeFilters(filters: SearchFilters, filterOptions: SearchFilters): SearchFilters {
```
esotrope: this was "sortFilters" and is now "normalizeFilters" 
this method no longer sorts in-place and returns sorted copies of the arrays we want to sort (for immutability principles)

webapp/src/app/data/search/search.service.ts
```diff
@@ -1,27 +1,108 @@
...
+    grades: toArraySet(filters.grades || [])
+      .sort(byRanking(filterOptions.grades)),
+    subjects: toArraySet(filters.subjects || [])
+      .sort(byRanking(filterOptions.subjects)),
+    claims: toArraySet(filters.claims || [])
+      .sort(byCode),
+    targets: toArraySet(filters.targets || [])
+      .sort(byCode),
+    standards: toArraySet(filters.standards || [])
+      .sort(byCode)
```
esotrope: idea here for using `sort(byRanking(originalOptions.whatever))` is so that we can respect the backend natural order of options provided. I understand these are hard coded at the moment but down the line the backend should be providing the natural order of its entities.
For the meantime I kept the claims, targets, standards sorted by code - felt that might actually make sense since there are so many.

webapp/src/app/data/search/search.service.ts
```diff
@@ -1,27 +1,108 @@
...
+  };
+}
+
+function mergeSearchFilters(filtersList: SearchFilters[], query: string): SearchFilters {
```
esotrope: just a rename.
Also note that I pulled some of the pure-functions up top.
when the function is a pure function there is no need to make it a private member - makes it more annoying to unit test if/when we want to do that.

webapp/src/app/data/search/search.service.ts
```diff
@@ -33,53 +114,60 @@ export class SearchService {
   getDefaultFilters(): Observable<SearchFilters> {
     // return this.dataService.get('/api/search/filters');
     // TODO: Replace with an actual way to get default filters
-
-    return of(JSON.parse(JSON.stringify(initialSearchFilters)));
+    return of(initialSearchFilters);
```
esotrope: was there some magic to this at some point?

webapp/src/app/data/search/search.service.ts
```diff
@@ -33,53 +114,60 @@ export class SearchService {
...
-        (acc, page) => {
-          acc.results.push(page['hydra:member'].map(this.resourceSummaryFromJson));
-          acc.filters.push(page['hydra:member'].map(this.extractFilters));
-          return acc;
-        },
-        { results: [], filters: [] }),
+  fetchAllResults(request: SearchRequestModel): Observable<SearchResults> {
+    return forkJoin(
+      this.getDefaultFilters(),
+      this.dataService.get('/api/search_dl_resources', request).pipe(
```
esotrope: composed this observable to depend on the result of the default options (since we will use them for sorting)

webapp/src/app/data/search/search.service.ts
```diff
@@ -206,67 +294,7 @@ export class SearchService {
...
-   */
-  private sortFilters(filters: SearchFilters) {
-
-    // Sort the filters. Here we will need to have logic specific to each
-    // filter type.
-    const sortOnCode = (a, b) => {
-      return a.code > b.code ? 1 :
-             a.code < b.code ? -1 :
-             0;
-    };
```
esotrope: this looks like a string sort so i replaced it with `String.localeCompare`

webapp/src/app/data/search/search.service.ts
```diff
@@ -206,67 +294,7 @@ export class SearchService {
...
-
-    const typesOrder = ['ir', 'cp', 'pl', 'fs', 'as'];
-
-    filters.grades.sort((a, b) => +a.code.substring(1) - +b.code.substring(1));
```
esotrope: this was the original problem i was addressing - not a safe assumption that the grade codes can be sorted this way.

webapp/src/app/data/search/search.service.ts
```diff
@@ -206,67 +294,7 @@ export class SearchService {
...
-    filters.subjects.sort(sortOnCode);
-    filters.claims.sort(sortOnCode);
-    filters.targets.sort(sortOnCode);
-    filters.standards.sort(sortOnCode);
-    filters.resourceTypes = typesOrder
-      .map(sortedCode => filters.resourceTypes.find(f => f.code === sortedCode))
```
esotrope: we were already using a similar idea to the "natural order" here

webapp/src/app/data/search/search.service.ts
jdbernard: The default filters are not a reliable source for sorting, unfortunately. The principle behind our search filtering logic is that the available filters are derived from the data set itself. For example, we don't encode the hierarchy of Grade/Claim/Target anywhere in our code. After a search we compose the set of filters that remain valid choices by inspecting the filter values present in the result set.

The reason this is relevant here is that we cannot depend on our mocked initial filter set to be accurate. As SB adds to the data set the universe of possible filter values may change (we currently have ~250 resources in the data set and expect a total of around ~900). For example, they will be adding resources for grade `Kindergarten` or for subject `Science`.

Really we need a new API that returns the universe of possible filter values with their preferred ordering instead of relying on an initial filter set that will go out of date whenever they add new resources. I'm hesitant to tie more functionality to the initial filter set.

webapp/src/app/data/search/search.service.ts
jdbernard: Yes. The search UI code later (particularly SearchComponent) may modify the SearchFilters object that it receives. The JSON parse/stringify was a way of doing a deep clone of the object so that our initial filter set doesn't get modified. In the long-term we want even this initial set to come from the API, so we'd get a fresh object after every call.

Ideally our SearchComponent would not modify it's inputs directly, but I haven't had time to refactor that component (so I've just been adding to the problem).

webapp/src/app/data/search/search.service.ts
jdbernard: Yes, and the bigger issue, I elaborate in a later comment, is that this search filter set is dynamic and data-driven. The filters available are calculated based on the values present in the data.

webapp/src/app/data/search/search.service.ts
jdbernard: In particular, please re-test the scenario from [DL-291](https://smarterbalanced.atlassian.net/browse/DL-291) to make sure this doesn't re-introduce this bug.

esotrope: good to know - thanks for the info.

esotrope:
After discussing offline I've decided to re-approach this in a less intrusive way : J

PR #169, opened 2020-01-28
fixes sorting of grade codes and adds sorting support methods

esotrope:


webapp/src/app/common/sorting/sorting.spec.ts
```diff
@@ -0,0 +1,47 @@
+import {byOrdering, byString, on} from './sorting';
+
+describe('byString', () => {
```
esotrope: coverage for sorting utils

webapp/src/app/common/sorting/sorting.ts
```diff
@@ -0,0 +1,125 @@
+/**
+ * Represents a javascript comparator passed to {@link Array#sort}
+ */
+export type Comparator<T> = (a: T, b: T) => number;
```
esotrope: sorting utils

webapp/src/app/common/utils.ts
```diff
@@ -20,31 +18,6 @@ export function getCssVar(cssVar: string, element?: HTMLElement) {
   return +getComputedStyle(element ? element : document.documentElement).getPropertyValue(cssVar).replace('px', '');
 }
 
-/**
- * See: https://jsperf.com/array-merge-performance
- */
-export function fastArrayMerge<T>(arrays: T[][]): T[] {
```
esotrope: i _think_ (and please correct me if i'm wrong) that this is flattening arrays 
e.g.
```[[1,2], [3,4]] -> [1,2,3,4]```
And i think i avoided it with a little code change i'll point out later

File: webapp/src/app/data/mock-data.ts [hunk 4]
```diff
@@ -113,8 +113,8 @@ export const initialSearchFilters: SearchFilters = {
     { code: 'ir', title: 'Instructional' },
     { code: 'cp', title: 'Connections Playlist' },
     { code: 'pl', title: 'Professional Learning' },
-    { code: 'as', title: 'Accessibility Strategy' },
-    { code: 'fs', title: 'Formative Assessment Strategy' }
+    { code: 'fs', title: 'Formative Assessment Strategy' },
+    { code: 'as', title: 'Accessibility Strategy' }
```
esotrope: made initial ordering reflect explicit ordering we use later

webapp/src/app/data/search/search.service.ts
```diff
@@ -56,30 +76,32 @@ export class SearchService {
...
       // collections once than to do it every time we map in another page of
       // results.
       reduce(
-        (acc, page) => {
-          acc.results.push(page['hydra:member'].map(this.resourceSummaryFromJson));
-          acc.filters.push(page['hydra:member'].map(this.extractFilters));
-          return acc;
+        (searchFiters, page) => {
+          searchFiters.results.push(...page['hydra:member'].map(this.resourceSummaryFromJson));
+          searchFiters.filters.push(...page['hydra:member'].map(this.extractFilters));
```
esotrope: using spread operator to "push all" and avoid the need to flatten the arrays later

webapp/src/app/search/search.component.ts
```diff
@@ -1,11 +1,20 @@
-import { AfterViewInit, Component, ElementRef, EventEmitter, Input, OnDestroy,
-  OnInit, Output, ViewChildren } from '@angular/core';
-import { ActivatedRoute, NavigationEnd, Router } from '@angular/router';
-import { MDCChipSet } from '@material/chips';
-import { Subscription } from 'rxjs';
-import { filter } from 'rxjs/internal/operators/filter';
-import { SearchFilters, emptyFilters } from '../data/search/search-filters.model';
-import { coalesce, whitelistKeys } from '../common/utils';
+import {
```
esotrope: removed unused imports

webapp/src/app/common/utils.ts
```diff
@@ -20,31 +18,6 @@ export function getCssVar(cssVar: string, element?: HTMLElement) {
...
-    for (let j = 0; j < arrays[i].length; j++) {
-      result[resultIdx] = arrays[i][j];
-      resultIdx++;
-    }
-  }
-  /*tslint:enable:prefer-for-of*/
-  return result;
-}
-
-export function makeQueryString(params: Params): string {
```
esotrope: this was unused

webapp/src/app/common/utils.ts
jdbernard: You are correct on both counts. However, the spread push syntax is several times slower (don't remember off the top of my head, but something like 8x slower).

In this case the API response is so slow I don't think the difference will matter.

esotrope: Yeah it's my experience JS processing time is negligible in the scheme of things so i tend toward ignoring it.

PR #170, opened 2020-01-29
fixes some interaction issues with teaser overlay

esotrope:
image

### DIFF
--- a/webapp/src/app/resource/components/preview-login-prompt/preview-login-prompt.component.ts
+++ b/webapp/src/app/resource/components/preview-login-prompt/preview-login-prompt.component.ts
@@ -1,19 +1,22 @@
-import { Component } from '@angular/core';
-import { Router } from '@angular/router';
-import { OktaAuthService } from '@okta/okta-angular';
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {OktaAuthService} from '@okta/okta-angular';
+import {BehaviorSubject} from 'rxjs';
 
 @Component({
   selector: 'sbdl-preview-login-prompt',
   templateUrl: './preview-login-prompt.component.html',
-  styleUrls: ['./preview-login-prompt.component.scss']
+  styleUrls: ['./preview-login-prompt.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class PreviewLoginPromptComponent {
-  loading = false;
+  loading$: BehaviorSubject<boolean> = new BehaviorSubject(false);
 
-  constructor(private oktaAuthService: OktaAuthService) {}
+  constructor(
+    private oktaAuthService: OktaAuthService
+  ) {}
 
-  login() {
-    this.loading = true;
-    this.oktaAuthService.loginRedirect();
+  onLoginButtonClick(): void {
+    this.loading$.next(true);
+    // this.oktaAuthService.loginRedirect();
   }
 }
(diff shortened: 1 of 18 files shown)


webapp/src/app/resource/instructional/content/instructional-content.component.html
```diff
@@ -3,28 +3,32 @@
   (sectionLoaded)="addDocumentSection($event)"
 ></sbdl-instructional-get-started>
 
-<sbdl-step-by-step
-  [steps]="resource.stepByStep"
-  (sectionLoaded)="addDocumentSection($event)"
-></sbdl-step-by-step>
+<sbdl-protected-content
+  [authenticated]="!resource.teaser"
+>
```
esotrope: new component to use to wrap protected content with

esotrope: also this is the only change in these -content components is to wrap the non-overview content with this

webapp/src/ui-kit/_mixins.scss
```diff
@@ -0,0 +1,9 @@
+@import "colors";
+
+@mixin window {
+  background-color: $white;
+  border-radius: 4px;
+  box-shadow: 0 4px 5px rgba(0, 0, 0, 0.2), 0 3px 14px rgba(0, 0, 0, 0.12), 0 8px 10px rgba(0, 0, 0, 0.14);
+  display: block;
+  padding: 1.5rem;
+}
```
esotrope: extracted from the login preview popover for potential shared use

webapp/src/app/resource/components/preview-login-prompt/preview-login-prompt.component.ts
```diff
@@ -1,19 +1,22 @@
...
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {OktaAuthService} from '@okta/okta-angular';
+import {BehaviorSubject} from 'rxjs';
 
 @Component({
   selector: 'sbdl-preview-login-prompt',
   templateUrl: './preview-login-prompt.component.html',
-  styleUrls: ['./preview-login-prompt.component.scss']
+  styleUrls: ['./preview-login-prompt.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
```
esotrope: this can be used on components that should only re-render when @Input() change or internal behavior subjects emit events

webapp/src/app/resource/components/protected-content/protected-content.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  styleUrls: ['./protected-content.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ProtectedContentComponent {
+
+  @Input()
+  authenticated: boolean;
+
+  @Input()
+  authorized: boolean;
```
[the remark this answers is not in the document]
jdbernard: It is possible for a user to be logged in and not have access to a resource, but in this case the resource is never made available for preview. I don't know that we will use this.

Resources can be:
1. Public
2. Private
3. Private + Tenant-specific.

Case 1 is always fully available.
Case 2 is shown in preview for anonymous users and in full for authenticated users.
Case 3 is always a hard deny (nothing shown) unless the user is both authenticated and belongs to that specific tenant.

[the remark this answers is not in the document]
esotrope: cool thanks for enumerating the cases 👍

PR #172, opened 2020-01-29
Dl 294 search

esotrope:


webapp/src/app/common/common.module.ts
```diff
@@ -1,59 +1,61 @@
...
-    ],
-    providers: [
-        DynamicHTMLRenderer,
-        { provide: DynamicHTMLOptions, useValue: dynamicOptions },
-        { provide: ANALYZE_FOR_ENTRY_COMPONENTS, useValue: dynamicOptions.components, multi: true },
-    ],
-    entryComponents: [ PopoverComponent ]
+  imports: [
+    CommonModule,
+    ReactiveFormsModule
```
esotrope: the added import - do you all have a shared IDE formatting config? my defaults are different than the existing.

webapp/src/app/common/controls/text-field/text-field.component.ts
```diff
@@ -1,12 +1,28 @@
...
   selector: 'sbdl-text-field',
   templateUrl: './text-field.component.html',
-  styleUrls: ['./text-field.component.scss']
+  styleUrls: ['./text-field.component.scss'],
+  providers: [
+    controlValueAccessorProvider(TextFieldComponent)
+  ]
 })
-export class TextFieldComponent implements OnInit, AfterViewInit {
+export class TextFieldComponent extends AbstractFormControlValueAccessor implements AfterViewInit {
```
esotrope: example of how to make a custom form-control that can be bound with `[(ngModel)]` or `formControlName` or `[formControl]` from the template - the issue was we had one-way binding set up on our query param so when we query the text field for its value later it didn't have it

webapp/src/app/common/controls/text-field/text-field.component.ts
```diff
@@ -28,16 +41,25 @@ export class TextFieldComponent implements OnInit, AfterViewInit {
...
+  ) {
+    super();
   }
-  
+
+  get control(): AbstractControl {
+    return this.controlContainer != null
+      ? this.controlContainer.control
+      : this._formControl;
+  }
```
esotrope: some ceremony to make this a robust custom form control

webapp/src/app/common/form/abstract-form-control-value-accessor.ts
```diff
@@ -0,0 +1,31 @@
+import {AbstractControl, ControlValueAccessor} from '@angular/forms';
+
+/**
+ * Extend this class to easily integrate your custom form control with angular's reactive forms
+ */
+export abstract class AbstractFormControlValueAccessor implements ControlValueAccessor {
```
esotrope: @jdbernard you probably vaguely remember this from clover - this makes its way into every angular project i work on - should be in their lib

webapp/src/app/search/results/search-results.component.html
```diff
@@ -6,7 +6,7 @@
     [numResults]="allResults.length"
     (goToResults)="scrollToResults()"
   ></sbdl-search>
-  <section #results class="search-results" width="120px" height="120px">
```
esotrope: unrelated cleanup

webapp/src/app/search/results/search-results.component.ts
```diff
@@ -21,12 +22,17 @@ export class SearchResultsComponent implements OnInit, OnDestroy {
 
   showAdvancedFiltersInitially: boolean;
 
-  private dataSubscription: Subscription;
-  private paramsSubscription: Subscription;
-  private routerSubscription: Subscription;
+  private readonly _destroyed$: Subject<void> = new Subject();
```
esotrope: unrelated cleanup - no need to track subscriptions and unsubscribe individually if you bind your observables with a `takeUntil(Subject)`

webapp/src/app/search/search.component.html
```diff
@@ -17,7 +17,7 @@ <h1 *ngIf="!newSearch" class="results-header">
   <span>
     Found {{ numResults }}
     <ng-container [ngPlural]="numResults" >
-      <ng-template ngPluralCase="=0">resources for "{{filters.query}}"</ng-template>
+      <ng-template ngPluralCase="=0">resources for "{{params.query}}"</ng-template>
```
esotrope: makes sure this doesn't update as you type

webapp/src/app/search/search.component.ts
```diff
@@ -124,8 +124,13 @@ export class SearchComponent implements  AfterViewInit, OnInit, OnDestroy {
...
 
   search(newParams: SearchQueryParams) {
-    const params = this.rectifyParams({...this.params, ...newParams });
-
-    this.router.navigate(['search', params]);
+    this.router.navigate([
+      'search',
+      this.rectifyParams({
+        ...this.params,
+        query: this.filters.query,
```
esotrope: makes sure to send the current search box state when a filter changes

webapp/src/app/search/search.module.ts
```diff
@@ -1,26 +1,30 @@
...
 
 @NgModule({
-  declarations: [
-    SearchComponent,
-    SearchResultsComponent,
-    SearchResultCardComponent
-  ],
   imports: [
     CommonModule,
+    FormsModule,
```
esotrope: the change

webapp/src/app/home/home.component.scss
```diff
@@ -183,3 +181,9 @@
     }
   }
 }
+
+.waterspot {
+  opacity: 0.7;
+  pointer-events: none;
+  user-select: none;
+}
```
esotrope: fixes issue with being able to select and drag waterspot images - maybe i could have just used div + css bg but this fixed it for now

webapp/src/app/common/common.module.ts
jdbernard: No. I've been relying on the tslint configuration to enforce a consistent style but it doesn't seem to care about this whitespace difference (whereas it does in other cases).

webapp/src/app/common/controls/text-field/text-field.component.html
```diff
@@ -1,27 +1,50 @@
...
   <sbdl-button
     class="primary"
-    (click)="onIconClick($event)"
-    [disabled]="disabled" >
-
-    <i *ngIf="fontAwesomeIcon" class="far {{fontAwesomeIcon}}" tabindex="0" role="button"></i>
+    (click)="onSubmitButtonClick()"
+    [disabled]="disabled"
+  >
+    <!-- TODO why not actually make this a button? -->
```
jdbernard: It is. :) `sbdl-button` internally uses a `<button>` element.

esotrope: Ah thx. i think my confusion arose with the `<i tabindex="0" role="button">`. Those attributes safe to remove now?

jdbernard:
Good question. My gut says yes, but double-check that the button still shows up in the tab-order if we do. I can’t remember if that code was part of the initial implementation, in which case it’s probably superfluous, or if it was added in a later WCAG review, in which case we’re probably doing something dumb like explicitly taking the `sbdl-button` out of the tab ordering.

PR #175, opened 2020-01-30
DL-319 Resource summary cards should render HTML-formatted summaries.

jdbernard:


webapp/src/app/resource/components/resource-summary-card/resource-summary-card.component.html
```diff
@@ -17,15 +17,15 @@
     </div>
 
     <div *ngSwitchCase="resourceTypes.ProfessionalLearning" class="details professional-learning caption emphasis-medium">
-      {{ resourceSummary.summary }}
+      <dynamic-html [content]="resourceSummary.summary"></dynamic-html>
```
esotrope: does this component do the html post-processing as well as rendering as html?
just wondering the diff between this and using `[innerHTML]`

jdbernard: It hooks into the Angular rendering pipeline to allow us to dynamically render Angular components (for example our `sbdl-tooltip` that we use for the embedded strategy links).

esotrope: interesting - i'm curious how you guys did that i'll have to take a peeksie

PR #176, opened 2020-01-30
DL-223 Initial implementation of user-based tenant theming.

jdbernard:


webapp/src/app/data/tenant-theme/tenant-theme.service.ts
```diff
@@ -0,0 +1,54 @@
...
+import { User } from 'src/app/data/user/user.model';
+
+const DEFAULT_TENANT_ID = 'default';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class TenantThemeService {
+
+  private tenantThemeConfig$ = new ReplaySubject<TenantThemeConfig>(1);
```
esotrope: _if_ we redesigned the user service to return a user object type: anonymous by default i think we could get away with:

```typescript
getTenantTheme(): Observable<TenantTheme> {
  return this.userService.user.pipe(
    mergeMap(user => user.anonymous
      ? of(defaultTheme)
      :  this.http.get(AppConfig.settings.tenantConfigPath),
    shareReplay({ refCount: true }) // caching
    catchError(error => { ... })
  )
}
```

jdbernard: I think we could probably get away with it now, just change to

```typescript
mergeMap(user => user ? of(defaultTheme) : ...```

Thoughts?

PR #182, opened 2020-01-30
 DL-64 Graceful error pages. 

jdbernard:
This is not entirely finished, but pushing so Nohemi can test as much as possible.

### DIFF
--- a/webapp/src/app/common/error-type.enum.ts
+++ b/webapp/src/app/common/error-type.enum.ts
@@ -1,0 +1,36 @@
+/*tslint:disable:max-line-length*/
+export enum ErrorType {
+  AuthNoAppAccess = 'auth-no-app-access',
+  ResourceIsPrivate = 'resource-is-private',
+  ResourceIsTenantSpecific = 'resource-is-tenant-specific',
+  Unknown = 'unknown'
+}
+
+export interface ErrorMessage {
+  title: string;
+  message: string;
+  details?: string;
+}
+
+export const errorMessages: Map<ErrorType, ErrorMessage>  = new Map([
+  [ErrorType.AuthNoAppAccess, {
+    title: 'Unable to access Tools for Teachers.',
+    message: 'It looks like your user is not configured with access to the Tools for Teachers application. Please contact help@smarterbalanced.org for assisstance.',
+  }],
+
+  [ErrorType.ResourceIsPrivate, {
+    title: 'Please login to check for access.',
+    message: 'The resource you\'re trying to view hs limited availability. Please login to see if your state/territory has access.',
+    details: 'The requested resource exists but is restricted by tenant the the current user is not authenticated.'
+  }],
+
+  [ErrorType.ResourceIsTenantSpecific, {
+    title: 'This resource is restricted.',
+    message: 'Unfortunately this resource is not available to educators in your state/territory.',
+    details: 'The requested resource exists but is restricted by tenant and the current user is not a member of the allowed tenants.'
+  }],
+  [ErrorType.Unknown, {
+    title: 'This resource is unavaiable.',
+    message: 'We apologize for the inconvenience.'
+  }]
+]);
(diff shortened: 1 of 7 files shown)


webapp/src/app/layout/login-callback/login-callback.component.ts
```diff
@@ -3,7 +3,15 @@ import { AfterViewInit, Component, Inject, OnInit } from '@angular/core';
 import { ActivatedRoute, NavigationExtras, Router } from '@angular/router';
 import { OktaAuthService } from '@okta/okta-angular';
 import { takeLast, takeWhile } from 'rxjs/operators';
+import { User } from 'src/app/data/user/user.model';
 import { UserService } from 'src/app/data/user/user.service';
+import { ErrorType } from 'src/app/common/error-type.enum';
+import { ERROR_PATH } from 'src/app/common/constants';
+
+interface ErrorWithDetails {
+  type: ErrorType;
```
esotrope: I'd be tempted to call this `ApplicationError` or `DigitalLibraryError` > `ErrorType`

jdbernard: Yeah, I'm also not even sure I like the structure as-is. Further feedback would be appreciated. I'll be revisiting this later tonight. Need to, for example, tie these into the resource request path.

webapp/src/app/common/error-type.enum.ts
```diff
@@ -0,0 +1,36 @@
+/*tslint:disable:max-line-length*/
+export enum ErrorType {
+  AuthNoAppAccess = 'auth-no-app-access',
+  ResourceIsPrivate = 'resource-is-private',
+  ResourceIsTenantSpecific = 'resource-is-tenant-specific',
```
esotrope: usually it is considered a security issue to leak this kind of granular info and its hidden behind a 404 not found error message. If the client wants it this way then it is up to them i guess though.

webapp/src/app/layout/error/error.component.ts
```diff
@@ -1,37 +1,57 @@
...
+    console.log('Error page params: ', this.route.snapshot);
 
-  ngOnDestroy() {
-    if (this.routerSubscription) {
-      this.routerSubscription.unsubscribe();
-      this.routerSubscription = undefined;
+    this.error = this.route.snapshot.params.type || ErrorType.Unknown;
+    if (!Object.values(ErrorType).some(v => v === this.error)) {
+      this.error = ErrorType.Unknown;
     }
```
esotrope: can a whole error fit in the params? - wondering if we shouldn't just pass the string ID of the error in the params

webapp/src/app/common/error-type.enum.ts
jdbernard: Agreed and yes, in this case the client has asked us specifically to make this information available to the user. From a security standpoint the risk is mitigated by the fact that the API is essentially read-only to begin with, and even with full read access there is nothing truly sensitive (the user just gets content for free).

webapp/src/app/layout/error/error.component.ts
```diff
@@ -1,37 +1,57 @@
...
+
+    this.errorMessage = errorMessages.get(this.error);
+    this.errorDetails = this.route.snapshot.params.details;
```
esotrope: instead of using snapshot let's use the reactive way:
```
this.route.params.pipe(map(x => x.details))
```

webapp/src/app/layout/error/error.component.ts
jdbernard: So the idea is that we're passing the string ID and an optional specific detailed error message. The full error is looked up from the Map defined alongside the enum. The optional specific detailed message would be used for short descriptions that are specific to this instance of the error.

jdbernard:
Merging so Noheni can test, but further feedback is still appreciated.

PR #187, opened 2020-01-31
Change z-index of footer and nav so they stack properly.

jdbernard:


webapp/src/app/layout/footer/footer.component.scss
```diff
@@ -5,7 +5,7 @@
 :host {
   background: $neutral-50;
   color: $neutral-500;
-  z-index: 999;
```
jdbernard: @jmesches-fairway I know we just added this. Can we walk through and talk about why we felt like we needed 999? Ideally we would know the index of all components in our stacking contexts and be able to set a specific value rather than hit the max. If we do need to max this out because of some browser quirk in printing, can we set that as a print style? Looking for input. 🙂

webapp/src/app/layout/navigation/navigation.component.scss
```diff
@@ -24,8 +24,7 @@
 
   overflow: hidden;
 
-  // TODO why?
```
jdbernard: Because we need it to be rendered above pretty much anything else when it is allowed to slide out from the left. We don't push or reflow the page content, we just overlay it. Happy to discuss.

webapp/src/app/layout/navigation/navigation.component.scss
```diff
@@ -12,7 +12,7 @@
   }
 
   transition-duration: 0.2s;
-  transition-property: width, z-index;
```
jdbernard: I don't think we actually ever alter this property and even if we did I don't know why we'd need to animate the z-index transition, so I'm removing this.

webapp/src/app/layout/navigation/navigation.component.scss
esotrope: Thanks - just put the comment there to point out that we shouldn't need z-indexing to achieve  what we want and we should think about how to tackle it without it

jdbernard: What do you see as our options other than z-index? Also, why to avoid?

esotrope: Ah. maybe worth a 5min talk in a parking lot?

PR #188, opened 2020-01-31
Updates to DL-64: redirect to error pages in more cases.

jdbernard:


webapp/src/app/layout/app-container/app-container.component.scss
```diff
@@ -41,8 +41,8 @@
       left: 0;
       position: absolute;
       top: 0;
-      width:  calc(100vw - 64px);
-      z-index: 10;
+      width:  100vw;
+      z-index: 4;
```
jdbernard: Allow the left-hand navigation to overlay the route-loader. There are cases where the user may be navigating from the left-nav. It looks odd if the loader partially overlays the nav. The other option would be for the loader to completely overlay the nav, but having the nav on top indicates to the user that we're not frozen, just loading content.

jdbernard:
Feedback and review still desired, merging to test on dev and make available for Nohemi to test.

webapp/src/app/common/global-error-handler.ts
```diff
@@ -19,10 +21,18 @@ export class GlobalErrorHandler implements ErrorHandler {
       this.logger = this.injector.get(LoggingService);
     }
 
+    if (!this.errorService) {
+      this.errorService = this.injector.get(TftErrorService);
+    }
```
esotrope: just curious - why do we lazy-get these services from the injector?

webapp/src/app/data/resource/resource.service.ts
```diff
@@ -59,6 +70,28 @@ export class ResourceService {
...
+        }
+      } else {
+        errorType = TftErrorType.ResourceUnavailable;
+      }
+
+      this.errorService.redirectTftError({ type: errorType, details: error.error });
+    });
+
+    return throwError(error);
+  }
```
esotrope: you could set this up to not need the subscribe like this i think:
```
return this.userService.user.pipe(
  tap(user => {
    this.errorService.redirectTftError({ type: getErrorType(user), ... });
    throw error;
  })
);
```

webapp/src/app/data/user/user.service.ts
```diff
@@ -87,6 +87,23 @@ export class UserService {
...
+  /**
+   * Convenience method to wrap the common case where a reactive pipeline
+   * should be executed only if a valid user session is present, returning a
+   * given default value otherwise.
+   */
+  withUserOrDefault<T>(fun: OperatorFunction<User, T>, defaultValue: T): Observable<T> {
+    return this.user$.pipe(
+      mergeMap(user => {
+        if (user) {
+          return fun(of(user));
```
esotrope: don't make fun of the user x P

webapp/src/app/layout/error/error.component.ts
```diff
@@ -16,18 +17,19 @@ import {UserService} from 'src/app/data/user/user.service';
 })
 export class ErrorComponent implements OnInit {
 
-  error: ErrorType;
-  errorDetails: string;
-  errorMessage: ErrorMessage;
+  error$: Observable<TftErrorType>;
+  errorDetails$: Observable<string>;
+  errorMessage$: Observable<TftErrorMessage>;
   loading$: BehaviorSubject<boolean> = new BehaviorSubject(false);
```
esotrope: awesome 👍

webapp/src/app/common/tft-error.service.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.router.events
+      .pipe(filter(evt => evt instanceof NavigationStart))
+      .subscribe((evt: NavigationStart) => this.lastTargetUrl = evt.url);
+    this.lastTargetUrl = location.path();
+  }
+
+  public redirectTftError(error: TftError) {
+    const destUrl = this.lastTargetUrl; // cache before we trigger another navigation
+    this.router.navigate([ERROR_PATH, error], { skipLocationChange: true })
+      .then(() => this.location.go(destUrl));
```
esotrope: not sure i follow this - i'll ask offline though

webapp/src/app/common/global-error-handler.ts
jdbernard: There's a comment in the source that explains (not visible in the diff). Basically this class is provided via injection for the ErrorHandler class. This gets pulled in before any of the other configuration (even AppConfig) so we can't depend on these to be present at the time of instantiation.

esotrope: ah thanks!

PR #189, opened 2020-01-31
Add OktaAuthModule to LayoutModule imports so OktaAuthService lives i…

jdbernard:
…n a proper zone.

### DIFF
--- a/webapp/src/app/layout/logout/logout.component.ts
+++ b/webapp/src/app/layout/logout/logout.component.ts
@@ -1,24 +1,28 @@
+import { Inject } from '@angular/core';
 import { APP_BASE_HREF } from '@angular/common';
-import { AfterViewInit, Component, OnInit } from '@angular/core';
+import { AfterViewInit, Component } from '@angular/core';
 import { Router } from '@angular/router';
 import { OktaAuthService } from '@okta/okta-angular';
+import { catchError, delay } from 'rxjs/operators';
 
 @Component({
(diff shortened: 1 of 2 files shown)


webapp/src/app/layout/logout/logout.component.ts
```diff
@@ -1,24 +1,28 @@
...
     private oktaAuthService: OktaAuthService,
     private router: Router
   ) {}
 
-  ngOnInit() {
-    this.oktaAuthService.logout();
-  }
-
   ngAfterViewInit() {
-    setTimeout(() => this.router.navigate([APP_BASE_HREF]), 3000);
```
jdbernard: Also, clearly this line was written while sleep-deprived.

webapp/src/app/layout/logout/logout.component.ts
```diff
@@ -1,24 +1,28 @@
...
+    const pauseAndRedirect = () => setTimeout(() => this.router.navigate([this.appBaseHref]), 2000);
+    this.oktaAuthService.logout().then(
+      pauseAndRedirect,
+      pauseAndRedirect);
```
esotrope: what is the artificial pause for?

jdbernard: Solely so the user has a chance to read the text on the page and doesn't feel like they missed something that may or may not have been important.

PR #196, opened 2020-02-03
DL-125 Add `contactUri` to the tenant theme configuration.

jdbernard:


webapp/src/app/home/home.component.html
```diff
@@ -5,7 +5,7 @@
 </div>
 <main id="main">
   <img class="logo-full"
-       [src]="(logo$ | async) | safeUrl"
+       [src]="(logo$ | async) | trustUrl"
```
esotrope: 👍

webapp/src/assets/config/tenant.config.json
```diff
@@ -1,12 +1,14 @@
 {
   "default": {
+    "contactUri": "http://www.smarterbalanced.org/contact/",
```
esotrope: feels like this could be put in the defaults and the tenant-specific config should just be overlaid on top of the defaults - basically a config hierarchy

jdbernard: True, but there really aren't any values for which we expect to typically be using defaults. Currently this is true for this value, but we're expecting them to provide contactUri values for all tenants. Felt like overkill to implement the logic for field-level fallback/hierarchy when I don't think we have any fields we expect to need it.

PR #202, opened 2020-02-03
Dl 276 playlist links

esotrope:


webapp/src/app/resource/playlist/resources/playlist-resources.component.html
```diff
@@ -1,5 +1,5 @@
 <h3>Playlist Resources</h3>
-<div class="caption emphasis-mediu">Quick links to the referenced materials.</div>
+<p class="caption emphasis-medium">Quick links to the referenced materials.</p>
```
esotrope: unrelated style consistency fix

jdbernard:
Actually, looks like the broken build may be something unrelated to this change. I think if you pull in the latest from dev first you'll pick up a fix for this (and the other failing build). We can discuss at standup.

PR #207, opened 2020-02-04
DL-382 Allow users without DL_EndUser the same access as anonymous us…

jdbernard:
…ers.


webapp/src/app/common/tft-error-type.enum.ts
```diff
@@ -20,7 +20,7 @@ export interface TftErrorMessage {
 export const errorMessages: Map<TftErrorType, TftErrorMessage>  = new Map([
   [TftErrorType.AuthNoAppAccess, {
     title: 'Unable to access Tools for Teachers.',
-    message: 'It looks like your user is not configured with access to the Tools for Teachers application. Please contact help@smarterbalanced.org for assisstance.'
+    message: 'It looks like your user is not configured with access to the Tools for Teachers application. Please contact support@smarterbalanced.org for assistance.'
```
esotrope: Should we say "full access" instead of access since they do have partial access?

PR #208, opened 2020-02-05
adds highlighting to table rows

esotrope:
not super pleased with this but it worked.

### DIFF
--- a/webapp/src/app/resource/components/outline/outline.component.ts
+++ b/webapp/src/app/resource/components/outline/outline.component.ts
@@ -1,11 +1,20 @@
-import { Component, OnInit, Input, HostListener } from '@angular/core';
-import { Location } from '@angular/common';
-import { Map } from 'immutable';
-import { Resource } from 'src/app/data/resource/model/resource.model';
-import { getCssVar } from 'src/app/common/utils';
-import { ResourceType } from 'src/app/data/resource/model/resource-type.enum';
-import { commentsSectionOptions } from '../simple-section/section.definitions';
-import { DocumentOutline, DocumentSection, DocumentSectionType as DST} from './document-outline.model';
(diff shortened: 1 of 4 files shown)


webapp/src/app/resource/resource.component.scss
```diff
@@ -146,21 +147,21 @@
...
+          }
+        }
+
+        tr.highlighted {
+          background-color: rgba($action-100, 0.4);
+          border-left: 4px solid rgba($action-500, 0.4);
+          border-radius: 0.25rem;
+          > th {
+            @include highlighted(0, -100%);
+          }
```
esotrope: so in totality, the `tr` is highlighted and we use a "clip/snippet" of the before pseudo element currently used for highlights.
I tried applying the current styles to `tr`  but it broke the table structure : /

webapp/src/app/resource/components/outline/outline.component.ts
```diff
@@ -1,11 +1,20 @@
...
+
+/**
+ * Gets the appropriate element to highlight given the selected section element
+ * @param element The section element
+ */
+function getHighlightTarget(element: HTMLElement): HTMLElement {
+  return element.closest('tr') != null
+    ? element.closest('tr')
+    : element;
+}
```
esotrope: this felt like an ok generalization - i see in the code there is a way using `#sectionHeader` to select things as the section element that are not in-fact headers though - this was slightly complicated by the fact that the tr was outside of the element declaring the header : /

jdbernard: `#sectionHeader` could be renamed. It's purpose is primarily to be a reference to the element that should be targeted for this kind of purpose. I'm good with this change though.

PR #209, opened 2020-02-05
adds notes to bottom of printed resources

esotrope:


webapp/src/app/resource/instructional/instructional-resource.component.html
```diff
@@ -47,21 +47,29 @@
...
+    ></sbdl-instructional-playlists>
   </sbdl-resource-properties>
+</aside>
+<aside
+  class="resource-notes"
+  [ngClass]="{
+    hidden: !notesVisible,
+    'has-notes': notes.length
+  }"
+>
```
esotrope: the hack was to make two aside elements and only display one at a time in "@media screen"
And display both in "@media print" and assign each to a different/new grid area in the template

webapp/src/app/resource/playlist/playlist-resource.component.html
```diff
@@ -44,16 +44,23 @@
...
+  >
     <sbdl-educational-details [properties]="resource.properties"></sbdl-educational-details>
     <sbdl-playlist-resources [topics]="resource.topics"></sbdl-playlist-resources>
   </sbdl-resource-properties>
+</aside>
+<aside
+  class="resource-notes"
+  [ngClass]="{
+    hidden: !notesVisible,
+    'has-notes': notes.length
```
esotrope: only show notes in the printed version if there are notes

webapp/src/app/resource/resource.component.scss
```diff
@@ -3,222 +3,228 @@
...
+    background-attachment: scroll;
+    background-position: right 184px bottom, bottom right;
+    min-height: 100vh;
+    background-color: $white;
+    overflow-x: hidden;
+
+
+    &.hidden {
+      display: none;
     }
```
esotrope: i was annoying and formatted this file - the change was to add .hidden styles and make sure the new `.resource-notes` class has the same styles as the `.resource-properties` class

PR #211, opened 2020-02-06
applies consistent aria-hidden to non-screen read images

esotrope:


webapp/src/app/bookmarks/bookmark-list/bookmark-list.component.html
```diff
@@ -4,7 +4,7 @@
 </nav>
 <main id="main" class="bookmark-list">
   <div class="top-header" aria-hidden="true">
-    <img src="/assets/images/yellow-spot.png"/>
+    <img src="/assets/images/yellow-spot.png" aria-hidden="true"/>
```
esotrope: this was probably the only real offender. the rest i replaced `alt=""` with `aria-hidden="true"` because it felt more explicit in intent

PR #212, opened 2020-02-06
Dl 389 wcag concerns

esotrope:
TODOs:
1. tabbing inside of attachment modal
2. tabbing position after closing modal

### DIFF
--- a/webapp/src/app/resource/components/header/header.component.html
+++ b/webapp/src/app/resource/components/header/header.component.html
@@ -1,5 +1,5 @@
 <div class="resource-type">{{ resource.type | resourceType }}</div>
-<h1 *ngIf="!showIconsCol; else titleWithIcons">{{ properties.title }}</h1>
+<h1 *ngIf="!showIconsCol; else titleWithIcons"><a [routerLink]="['/resource', resource.id]">{{ properties.title }}</a></h1>
 <ng-template #titleWithIcons>
   <div class="title-with-icons">
     <div class="icon-col">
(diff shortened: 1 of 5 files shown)


webapp/src/app/resource/components/header/header.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="resource-type">{{ resource.type | resourceType }}</div>
-<h1 *ngIf="!showIconsCol; else titleWithIcons"><a [routerLink]="['/resource', resource.id]">{{ properties.title }}</a></h1>
+<h1 *ngIf="!showIconsCol; else titleWithIcons">{{ properties.title }}</h1>
```
esotrope: was there a reason the resource heading was a link to itself? felt like noise.

webapp/src/app/resource/components/outline/outline.component.ts
```diff
@@ -66,7 +66,7 @@ export class OutlineComponent implements OnInit {
   scrollTo(section: DocumentSection): void {
     section.elementRef.scrollIntoView({behavior: 'smooth', block: 'start', inline: 'nearest'});
     section.elementRef.classList.add('highlighted');
-    this.location.go(this.location.path() + '#' + section.component.sectionId);
+    location.href = this.location.path() + '#' + section.component.sectionId;
```
esotrope: this actually changes the unseen tab cursor position after linking

webapp/src/styles.scss
```diff
@@ -21,6 +21,11 @@
     --nav-width: 331px;
 }
 
+html {
+  // enables smooth scrolling for traditional anchor links
+  scroll-behavior: smooth;
```
esotrope: needed for the smooth scrolling you guys are going for

webapp/src/app/resource/components/header/header.component.html
jdbernard: Yes. The idea was that if this gets exported/printed as a PDF we wanted to preserve links internal within the document, and wanted the title itself to be a link back to the resource on the site (in case people are working with the PDF).

esotrope:
@jdbernard they want it tab-accessible though? would seem annoying to tab to that as a keyboard user.

jdbernard:
@esotrope No, I don't think it needs to be in the tab order.

PR #218, opened 2020-02-06
DL-351: fixed merge issues

jmesches-fairway:


webapp/src/app/resource/components/attachments/card/attachment-card.component.scss
```diff
@@ -6,6 +6,16 @@
...
 :host {
 
+  a { 
+    text-decoration: none;
+
+    > .detail-info {
+      & > span { 
+        text-decoration: underline; 
+      }
+    }
```
esotrope: adding underline to something other than the anchor tag itself feels dubious but i'll assume it was needed.

nit: `& > span` can be simplified to `> span`

webapp/src/app/resource/components/resource-properties/resource-properties.component.scss
```diff
@@ -25,6 +25,11 @@
         height: calc(100% - 2rem); // $spacer * 2
         padding: $spacer;
       }
+
+      img { 
+        height: 279px;
+        width: 279px;
+      }
```
esotrope: should we extract a `$thumbnail-size` var here?

PR #222, opened 2020-02-06
Change global fragment scrolling behavior to only scroll if an elemen…

jdbernard:
…t exists for the given id.

### DIFF
--- a/webapp/src/app/layout/app-container/app-container.component.ts
+++ b/webapp/src/app/layout/app-container/app-container.component.ts
@@ -42,7 +42,8 @@ export class AppContainerComponent implements OnInit {
     this.route.fragment.subscribe((fragment) => {
       if (fragment && !this.router.url.includes(OKTA_CALLBACK_PATH)) {
         setTimeout(() => {
-          document.querySelector('#' + fragment).scrollIntoView({behavior: 'smooth'});
+          const el = document.querySelector('#' + fragment);
+          if (el) { el.scrollIntoView({behavior: 'smooth'}); }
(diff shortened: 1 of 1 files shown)


webapp/src/app/layout/app-container/app-container.component.ts
```diff
@@ -42,7 +42,8 @@ export class AppContainerComponent implements OnInit {
     this.route.fragment.subscribe((fragment) => {
       if (fragment && !this.router.url.includes(OKTA_CALLBACK_PATH)) {
         setTimeout(() => {
-          document.querySelector('#' + fragment).scrollIntoView({behavior: 'smooth'});
+          const el = document.querySelector('#' + fragment);
+          if (el) { el.scrollIntoView({behavior: 'smooth'}); }
         });
       }
     });
```
esotrope: curious why this isn't the default behavior - the browser usually does this for you. i also noticed that hashes get stripped (at least on the resource page) after initial navigation

jdbernard: You mean the navigation to the element with the matching ID? I believe it's because the elements referenced in the URL don't exist when the page is first loaded and the browser behavior would be triggered.

Angular provides this functionality for top-level components (see https://angular.io/api/router/ExtraOptions#anchorScrolling) but this doesn't currently work for child components, so we have to roll our own.

esotrope: interesting - thanks for the explanation : j

PR #223, opened 2020-02-06
DL-392 Use div's to contain Font Awesome icons that are conditionally…

jdbernard:
… displayed.

### DIFF
--- a/webapp/src/app/bookmarks/bookmark-actions/bookmark-actions.component.html
+++ b/webapp/src/app/bookmarks/bookmark-actions/bookmark-actions.component.html
@@ -10,17 +10,15 @@
   "
   (click)="toggleBookmarked()"
 >
-  <ng-container *ngIf="updatingBookmarked; else notUpdating">
+  <div *ngIf="updatingBookmarked">
     <i class="fas fa-spinner fa-spin"></i>
-  </ng-container>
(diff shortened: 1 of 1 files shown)


webapp/src/app/bookmarks/bookmark-actions/bookmark-actions.component.html
```diff
@@ -10,17 +10,15 @@
...
-      <i class="far fa-bookmark"></i>
-    </ng-template>
-  </ng-template>
+  </div>
+  <div *ngIf="!updatingBookmarked && bookmark">
+    <i class="fas fa-bookmark"></i>
+  </div>
+  <div *ngIf="!updatingBookmarked && !bookmark">
+    <i class="far fa-bookmark"></i>
+  </div>
```
jdbernard: Font Awesome does some of its own element replacement magic to turn these `<i>` elements into `<svg>`s. This creates new elements that live outside the knowledge of the `ng-template` or `ng-containers` that originally contained them. To avoid this we have to wrap them in concrete elements and attach the logic to the wrapping element.

esotrope: interesting. when did font awesome start doing that? i thought it was font-based svg

PR #227, opened 2020-02-07
fixes focus issues with preview modal

esotrope:


webapp/src/app/resource/components/attachments/preview/preview.component.html
```diff
@@ -1,6 +1,14 @@
-<div class="preview-container">
-  <sbdl-button class="plain close" #closeButton (click)="onClose.emit(true)">
-    Close <sbdl-icon icon="circle-close"></sbdl-icon>
+<div
+  class="preview-container"
+  cdkTrapFocus
```
esotrope: prevents focus from escaping the modal when it is open

webapp/src/app/resource/components/attachments/preview/preview.component.ts
```diff
@@ -48,14 +49,26 @@ export class PreviewComponent implements AfterViewInit, OnDestroy, OnInit {
...
+
+  onDownloadButtonClick(attachment: ResourceAttachment): void {
     this.attachmentsService.download(attachment);
   }
+
+  private close(): void {
+    this.onClose.emit(true);
+    if (this.previousActiveElement != null) {
+      this.previousActiveElement.focus();
+    }
```
esotrope: main fix is to track the active element when opening the modal then re-focus it when closing

PR #233, opened 2020-04-01
TFT-76

steve-nystrom:
Updated 'browse for resources' labels to 'filter resources' on the main search page

### DIFF
--- a/webapp/src/app/search/search.component.html
+++ b/webapp/src/app/search/search.component.html
@@ -8,11 +8,8 @@
 >
 </sbdl-text-field>
 
-<div *ngIf="newSearch" class="caption browse-caption">
+<div class="caption">
   Start typing a topic, claim, target, or <a href="javascript:void(0)" (click)="showAdvanced = true">filter resources</a>.
-</div>
-<div *ngIf="!newSearch && numResults > 0" class="caption filters-caption">
-  Start typing a topic, claim, target, or <a href="javascript:void(0);" (click)="showAdvanced = true">filter resources</a>.
 </div>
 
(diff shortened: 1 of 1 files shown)


webapp/src/app/search/search.component.html
```diff
@@ -9,10 +9,10 @@ <h1 *ngIf="showHeadings && newSearch" class="new-search">Search for Resources </
 </sbdl-text-field>
 
 <div *ngIf="newSearch" class="caption browse-caption">
-  Start typing a topic, claim, target, or <a href="javascript:void(0)" (click)="showAdvanced = true">browse for resources</a>.
+  Start typing a topic, claim, target, or <a href="javascript:void(0)" (click)="showAdvanced = true">filter resources</a>.
 </div>
 <div *ngIf="!newSearch && numResults > 0" class="caption filters-caption">
-  Start typing a topic, claim, target, or <a href="javascript:void(0);" (click)="showAdvanced = true">advanced search</a>.
+  Start typing a topic, claim, target, or <a href="javascript:void(0);" (click)="showAdvanced = true">filter resources</a>.
 </div>
```
jdbernard, line 16: Now that the content of these two is exactly the same, can we consolidate the `*ngIf` logic and collapse it down to one? Also, I don't think the `browse-caption` and `filter-caption` are actually doing anything. Can we remove them?

steve-nystrom: Let me take a quick look and make sure those classes aren't being used.  That was the only reason I didn't consolidate it down to one line.

steve-nystrom: Consolidated the *ngIf logic down to one.

PR #238, opened 2020-04-15
Feature/T4T-198

steve-nystrom:


webapp/src/app/search/search.component.ts
```diff
@@ -123,14 +174,65 @@ export class SearchComponent implements  AfterViewInit, OnInit, OnDestroy {
...
+
+    this.popover.onClose.subscribe(this.close);
+    this.storageService.setSearchLoginWarningDisplayed(true);
+  }
+
+  private shouldDisplayLoginWarning(): boolean {
+    const previouslyDisplayed = this.storageService.getSearchLoginWarningDisplayed();
+    const displayPopover = !this.authenticated && !previouslyDisplayed;
+
+    return displayPopover;
```
jdbernard, line 236: Have we thought about encapsulating this functionality within the login warning component itself? If we re-use this login component in other places how much of this code will be have to duplicate in those places?

PR #239, opened 2020-04-20
T4T-200

steve-nystrom:
Modified the logic for displaying the warning login popover so it is not displayed when the user has selected accessibility strategies as one of the search criteria filters.


jdbernard:
This shares a number of duplicate commits with dev. Let's close this PR and create a new branch with just the relevant changes.

PR #240, opened 2020-04-20
Feature/T4T-199

steve-nystrom:
Display the login warning popover if the users clicks on the filter resources link and all other conditions have been met to display the popover.


steve-nystrom:
Relevant changes were put into brand 199-1.

PR #245, opened 2020-04-24
Feature/T4T-35

steve-nystrom:
Allow Users to edit their resource notes. Branch was created off of the T4T-34 branch.


jdbernard:
Looking at the merge, I think we want both changes.

PR #247, opened 2020-04-29
T4T-200

steve-nystrom:
Corrected the following issues:
  - Login warning popup was being displayed for searches within the Accessibility Strategy filter page
  - When redirected after authentication to the formative strategy page, 'resourceTypes=fs' was not included in the redirect path

### DIFF
--- a/webapp/src/app/search/login-warning/login-warning.component.ts
+++ b/webapp/src/app/search/login-warning/login-warning.component.ts
@@ -27,6 +27,7 @@ export class LoginWarningComponent {
   constructor(
     private oktaAuthService: OktaAuthService,
     private router: Router,
+    private route: ActivatedRoute,
     private loginWarningService: LoginWarningService,
     private userService: UserService,
     private storageService: StorageService,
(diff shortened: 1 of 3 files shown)


webapp/src/app/search/login-warning/login-warning.component.ts
```diff
@@ -46,7 +46,7 @@ export class LoginWarningComponent {
   }
 
   login() {
-    const redirectUrl = this.router.createUrlTree(['/search'], { queryParams: this.queryParams }).toString().replace('search?', 'search;');
+    const redirectUrl = this.router.createUrlTree([], { queryParams: this.queryParams, relativeTo: <ActivatedRoute><unknown>this.router.url }).toString().replace('?',';');
```
jdbernard, line 49: Is this a type coercion? If so, can we sync briefly on why this is necessary?

jdbernard: Per our discussion, let's use an injected ActivatedRoute:

```typescript
import {ActivatedRoute} from '@angular/router';
constructor(private route: ActivatedRoute) {}
// then instead of <ActivatedRoute>...
relativeTo: this.route
```

PR #248, opened 2020-04-29
Feature/T4T-217

steve-nystrom:
Created the new login page and modified all login points to navigate to the new page.


webapp/src/app/layout/login/login.component.html
```diff
@@ -0,0 +1,20 @@
...
+<main id="main">
+  <img class="logo-full"
+    [src]="(logo$ | async)"
+    alt="Tools for Teachers from Smarter Balanced">
+
+  <h1>Select your location to login.</h1>
+
+  <div class="body">
+    <p>Access to the content on this site is available to educators in subscribing states, territories, and districts.
+      Make a selection below to login.
```
jdbernard, line 15: Where is the selection UI to which this refers?

webapp/src/app/layout/login/login.component.html
```diff
@@ -0,0 +1,20 @@
...
+    </p>
+  </div>
+
+  <a href="#">Not in a subscribing location? Learn more about Tools for Teachers access.</a>
```
jdbernard, line 19: Is this a placeholder link? Do we have line-of-sight on what we want this to be?

webapp/src/app/layout/error/error.component.ts
```diff
@@ -57,6 +57,6 @@ export class ErrorComponent implements OnInit {
 
   onLoginButtonClick(): void {
     this.loading$.next(true);
-    this.oktaAuthService.loginRedirect(this.location.path());
+    this.router.navigate(['/auth/login']);
```
jdbernard, line 60: This change I expect, but where does the actual login functionality live now? There doesn't seem to be anywhere where we call `loginRedirect` anymore. Are we missing some commits?

webapp/src/app/layout/login/login.component.html
steve-nystrom: This was broken up into several different jira tickets. The selector was added in ticket T4T-218.

webapp/src/app/layout/error/error.component.ts
steve-nystrom: For this ticket I routed the user to the new login page.  Ticket T4T-220 addresses redirecting the user back to the page they were on before being routed to the login page.

webapp/src/app/layout/login/login.component.html
steve-nystrom: This is a placeholder. I need to find out where we need to redirect the user.

steve-nystrom:
These changes were included with the merge for T4T-220 so there is no need to merge this in as well.

PR #250, opened 2020-05-03
Wrap MathJax in a way we can call from Angular, then typeset resource…

jdbernard:
…s as they render.

### DIFF
--- a/webapp/src/app/common/mathjax.service.ts
+++ b/webapp/src/app/common/mathjax.service.ts
@@ -68,7 +68,10 @@ export class MathJaxService {
       this.scriptService.require('https://polyfill.io/v3/polyfill.min.js?features=es6'),
       this.scriptService.require('https://cdn.jsdelivr.net/npm/mathjax@3/es5/mml-svg.js')
     ).subscribe(_ => {
-      this.mathjax$.next((window as any).MathJax);
+      const mjObj = (window as any).MathJax;
+      if (mjObj) {
+        this.mathjax$.next(mjObj);
+      }
     });
   }
 
--- a/webapp/src/app/common/mathjax.service.ts
+++ b/webapp/src/app/common/mathjax.service.ts
@@ -40,30 +40,44 @@
  * loaded once and accessed via the global window object. In order to make that
  * play nice with Angular we are:
  *
- * 1. including the script from CDNJS via index.html, and *never* importing in
- *    TypeScript,
+ * 1. including the script from CDNJS via a <script> tag appended to the body
+ *    (via ScriptService), and *never* importing in TypeScript,
  * 2. creating a bare-bones interface in globals.d.ts to enable static type
  *    checking for our use-cases (and satisfying the TypeScript compiler),
- * 3. hiding/wrapping the global variable access behind a factory function, and
- * 4. using that factory function to expose the MathJax instance to the rest of
- *    our app in a standard manner (via Angular dependency injection).
+ * 3. hiding/wrapping the MathJax global object behind a service that wraps
+ *    calls in a way Angular is accustomed to receiving them.
  *
  * This way we should at least be able to write proper tests, mocking out
  * MathJax as needed.
  */
 
-import {InjectionToken} from '@angular/core';
+import {Injectable} from '@angular/core';
+import {concat, Observable, ReplaySubject} from 'rxjs';
+import {mergeMap} from 'rxjs/operators';
+import {ScriptService} from './script.service';
 
-// This is the injection token that will be used to ID the MathJax instance in
-// the Angular DI framework.
-export const MATHJAX_INST = new InjectionToken<MathJax>('MathJax');
+@Injectable({
+  providedIn: 'root'
+})
+export class MathJaxService {
+  private mathjax$ = new ReplaySubject<MathJax>(1);
+  private typesetPromise: Promise<void> = Promise.resolve(null);
 
-// This is the factory function that wraps the global window access and
-// provides the MathJax instance.
-export function getMathJax(): MathJax {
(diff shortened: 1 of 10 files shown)


webapp/src/index.html
```diff
@@ -12,7 +12,7 @@
 
     <!-- MathML Support via MathJax -->
     <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
-    <script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
+    <script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/mml-svg.js"></script>
```
esotrope, line 15: is mathml throughout the whole app in dynamic html? or just specific portions?
Asking because you may be able to lazy load the script in the component where you expect math ml.
Also, since this is async can we guarantee it is loaded by the time the component is loaded and calls getMathJax() ? i think loading it on demand you could ensure it is there by using 'script.load` callback.
In another SBAC project we are using mathjax as well. Let me know if you'd like me to go over it with you.

jdbernard: Re. being throughout the whole app, no but basically yes. It is only (currently) present in resource content, but it can be almost anywhere in resource content and resources make up the bulk of the app.

Re. guaranteeing that it loads in time, that's a good question. I'm relying on a fuzzy feeling of Angular waiting for window.onload (which I think includes async scripts) and the fact that Angular lazy-loads injectable things, expecting that `getMathJax()` won't be called until MathJax is present. However, that is like 4 layers of shaky assumptions so I'm open to other ideas.

Yes, if we have time I would like to look at other examples of how we've integrated this, particularly into an Angular project.

webapp/src/app/resource/resource-type-strategy.component.ts
```diff
@@ -99,11 +99,7 @@ export class ResourceTypeStrategyComponent implements AfterViewInit, OnInit, OnD
   }
 
   ngAfterViewInit() {
-    if (this.typesetPromise) {
-      this.typesetPromise = this.typesetPromise.then(this.mathjax.typesetPromise);
-    } else {
-      this.typesetPromise = this.mathjax.typesetPromise();
-    }
+    this.mathjaxService.typeset().pipe(take(1));
```
esotrope, line 102: because there is no `subscribe()` called here it tells me that the typeset() method has a side-effect during observable creation... taking a look at how we avoided this.

webapp/src/app/common/mathjax.service.ts
```diff
@@ -40,30 +40,44 @@
...
+
+  public typeset(): Observable<void> {
+    return this.mathjax$.pipe(
+      mergeMap(mj => {
+        this.typesetPromise = this.typesetPromise.then(mj.typesetPromise);
+        return this.typesetPromise;
+      })
+    );
   }
 }
```
esotrope, line 83: Here's the other sbac mathjax service. It wraps the promise.then() calls in a `new Observable()` to defer the action until subscribe() is called:
```typescript
@Injectable({
  providedIn: 'root',
})
export class MathJaxService {
  private _typesetPromise: Promise<void> = Promise.resolve();

  constructor(
    private readonly scriptService: ScriptService
  ) {}

  typesetPromise(elements?: Element[]): Observable<void> {
    return this.scriptService.require(
      'assets/vendor/verdana-mathjax/verdana-mathjax.min.js'
    ).pipe(
      map(() => (window as any).MathJax),
      flatMap(MathJax => new Observable<void>(observer => {
        // MathJax.typesetPromise() is async and uses shared state between method calls
        // This ensures that the method is not called until all previous calls complete
        // so that there are no errors or unexpected side-effects
        this._typesetPromise = this._typesetPromise
          .then(() => MathJax.typesetPromise(elements))
          .catch(error => {
            observer.error(error);
          })
          .then(() => {
            observer.next();
            observer.complete();
          });
      }))
    );
  }

}
```

jdbernard: Don't we already get that for free? If I add logs:

MathJaxService#typeset
```typescript
  public typeset(): Observable<void> {
    return this.mathjax$.pipe(
      mergeMap(mj => {
        console.log('thenning MathJax.typesetPromise');
        this.typesetPromise = this.typesetPromise.then(mj.typesetPromise);
        return this.typesetPromise;
      })
    );
  }
```

ResourceTypeStrategyComponent#ngOnInit
```typescript
  ngAfterViewInit() {
    console.log('get the MathJaxService.typeset Observable');
    const typesetObs = this.mathjaxService.typeset();

    console.log('waiting a second');
    setTimeout(() => {
      console.log('subscribing');
      typesetObs.pipe(take(1)).subscribe(() => {});
    }, 1000);
  }
```

Debug console log:
image

esotrope: Yeah, it will work either way - I just wanted to point out that the way it seems to be now breaks the observable contract in that the typesetting happens before .subscribe() is called. You could also just change the math jax service method signature to return a promise - that way it would be true to its contract

jdbernard: Can you double-check the above? I'm pretty sure based on the logs above that the underlying `MathJax.typesetPromise` function isn't being called until the observable is subscribed to. What am I missing?

webapp/src/app/resource/resource-type-strategy.component.ts
jdbernard: Good catch. Forgot to subscribe. 🤦

jdbernard:
Thanks again @esotrope and @bgsavage for your feedback!

PR #253, opened 2020-05-07
Feature/T4T-218

steve-nystrom:


steve-nystrom:
These changes were included with the merge for T4T-220 so there is no need to merge this in as well.

PR #258, opened 2020-05-11
Feature/T4T-220

steve-nystrom:


jdbernard:
@steve-nystrom Since this branch contains all the code from the T4T-219 branch, would it be easier to close that PR and just deal with this one?

steve-nystrom:
@jdbernard As you mentioned above, this feature branch contains all the code from the T4T-217, T4T-218, and T4T-219 branches (they all depended on the previous tickets features existing to expand the functionality of the new login page), so we should be able to just deal with this one.

PR #285, opened 2020-07-29
Adding IAIP Portal Link

JJ-Muller:
Enable access to IAIP via Tools for Teachers navigation bar based on authentication.

### DIFF
--- a/webapp/src/app/data/user/user.service.ts
+++ b/webapp/src/app/data/user/user.service.ts
@@ -71,7 +71,11 @@ export class UserService {
 
     sbacTenancyChain.some(chain => {
       if (chain.toLowerCase().includes('sb_iaip_user')) {
+        const segments = chain.split('|');
         iaip = true;
+        return true;
+    } else {
+      return false;
     }
   });
     return iaip;
--- a/webapp/src/app/data/user/user.service.ts
+++ b/webapp/src/app/data/user/user.service.ts
@@ -26,7 +26,7 @@ export class UserService {
   // produce a value. For example, see the DataService's usage.
   private user$: ReplaySubject<User> = new ReplaySubject(1);
   private hasOktaAuthToken$: ReplaySubject<boolean> = new ReplaySubject(1);
-  private hasIaipRole$: ReplaySubject<boolean> = new ReplaySubject(1); 
+  private hasIaipRole$: ReplaySubject<boolean> = new ReplaySubject(1);
 
   private jwtService: JwtHelperService;
 
@@ -72,9 +72,9 @@ export class UserService {
     sbacTenancyChain.some(chain => {
       if (chain.toLowerCase().includes('sb_iaip_user')) {
         iaip = true;
-    }     
+    }
   });
-    return iaip;  
+    return iaip;
   }
 
   static validateUserSession(user: User): TftError | null {
--- a/webapp/src/app/data/user/user.service.ts
+++ b/webapp/src/app/data/user/user.service.ts
@@ -68,12 +68,8 @@ export class UserService {
 
   static parseSbacTenancyChainForIAIP(sbacTenancyChain: string[]): boolean {
     return sbacTenancyChain.some(chain => {
-      if (chain.toLowerCase().includes('sb_iaip_user')) {
-        return true;
-    } else {
-      return false;
-    }
-  });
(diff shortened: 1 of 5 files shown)


webapp/src/app/data/user/user.service.ts
```diff
@@ -65,6 +66,21 @@ export class UserService {
     return validTenancies;
   }
 
+  static parseSbacTenancyChainForIAIP(sbacTenancyChain: string[]): boolean {
+    let iaip: boolean = false;
```
junyuchen1994, line 70: you don't need to specify the type for this let variable .

webapp/src/app/data/user/user.service.ts
```diff
@@ -65,6 +66,21 @@ export class UserService {
...
+
+    sbacTenancyChain.some(chain => {
```
junyuchen1994, line 72: To make the code look cleaner, you might want to refactor it to something like this.
 static parseSbacTenancyChainForIAIP(sbacTenancyChain: string[]): boolean {
    return sbacTenancyChain.some(chain => {
      if (chain.toLowerCase().includes('sb_iaip_user')) {
        const segments = chain.split('|');
        return true;
      } else {
        return false;
      }
    });
  }

webapp/src/app/data/user/user.service.ts
```diff
@@ -65,6 +66,21 @@ export class UserService {
...
+      if (chain.toLowerCase().includes('sb_iaip_user')) {
+        const segments = chain.split('|');
```
junyuchen1994, line 74: what are we doing with the constant segments? since the method is only returning a boolean value , this constant is never being used ,  do we want to do some additional error handling based on the constant value ?

webapp/src/app/data/user/user.service.ts
```diff
@@ -133,8 +153,10 @@ export class UserService {
   /**
    * Update our ReplaySubject with new User information depending on the new
    * auth state.
+   * Additionally update new iaipRole status
    */
   private updateUser = async (hasAuth) => {
+    let iaipCheck:boolean = false;
```
junyuchen1994, line 159: remove this variable declaration and replace it with const at line 173

webapp/src/app/data/user/user.service.ts
```diff
@@ -147,9 +169,15 @@ export class UserService {
         this.user$.next(null);
       } else {
         this.user$.next(await this.readUserFromOkta());
+        const userInfo = await this.oktaAuthService.getUser();
+        iaipCheck = UserService.parseSbacTenancyChainForIAIP(userInfo.sbacTenancyChain);
```
junyuchen1994, line 173: can you do 
const iaipCheck =UserService.parseSbacTenancyChainForIAIP(userInfo.sbacTenancyChain);
and remove  variable declaration

webapp/src/app/data/user/user.service.ts
```diff
@@ -147,9 +169,15 @@ export class UserService {
...
+        if(iaipCheck){
```
junyuchen1994, line 174: could you put space around the parentheses? when I checked out the code in my IDE , it's complaining about the spacing since angular checks for code formatting at compile time.

webapp/src/app/data/user/user.service.ts
JJ-Muller: you are right. that was actually an old code line that i no longer need.  i will remove.

webapp/src/app/data/user/user.service.ts
```diff
@@ -65,6 +66,16 @@ export class UserService {
     return validTenancies;
   }
 
+  static parseSbacTenancyChainForIAIP(sbacTenancyChain: string[]): boolean {
+    return sbacTenancyChain.some(chain => {
+      if (chain.toLowerCase().includes('sb_iaip_user')) {
```
junyuchen1994, line 71: i think it can be further optimized to be something like this
 return sbacTenancyChain.some(chain => {
     return  chain.toLowerCase().includes('sb_iaip_user');

since chain.inludes(' ') is already returning a boolean value , we can avoid the redundant if else statement.

test this in you local and let me know if it works for you.

JJ-Muller: i did not realize that, thank you!

PR #316, opened 2020-10-26
long footer text style fix.

jchen1005:


JJ-Muller (changes requested):
@jchen1005 looks good, but can we up the max-width to 18rem?

JJ-Muller (approved):

PR #317, opened 2020-10-30
t4t-496: handle 404 page not found.

jchen1005:


JJ-Muller (commented):
@jchen1005 is there a benefit of consolidating vs. not adding the page-not-found related code to the existing error.component.ts ?

jchen1005:
@JJ-Muller.  just saw this. this github account is linked to my old email , I don't usually check those github notifications. you can leave a message on jira if you have posted some comments for future PRs .
to answer your question ,  I've tried other options  but this is the only way that works . I don't see a issue adding a few line of code to reuse the existing error.component

JJ-Muller (commented):
@jchen1005 if you did add the lines to reuse the existing error.component would that erase the need for the page-not-found.component.ts file?

jchen1005:
@JJ-Muller  angular handles 404 error through the router , it scans all the registered router in the system , if the url user enter does not match the registered  route , it goes to Page Not Found  component , to be able to reuse the error.component.ts , we need to pass what type of error message to be display. that  why there is a extra page not found component as there is no way to pass a argument or object inside the router setting ts file

ramacsh (approved):

PR #336, opened 2021-02-11
Feature/T4T-639 handle parenthesis in strategy's title

jim-ruhl-accenture-com:
updated the class EmbedStrategyLinksService to encode for parenthesis when they are in the strategy's title.


JJ-Muller (changes requested):
requesting to make sure the space between the strategy icon and the title is retained.

JJ-Muller (approved):

PR #347, opened 2021-03-03
T4t-370 Adding the IAIP buttons

jim-ruhl-accenture-com:
A fix for Firefox and the change on the footer issue (T4T-642)


JJ-Muller (changes requested):
@jim-ruhl-accenture-com we do not want to change the centered layout placement of the header.

PR #364, opened 2021-04-09
T4T-734 - default captions to overlay

jim-ruhl-accenture-com:
made the default for captions to be overlaid.


JJ-Muller (changes requested):
@jim-ruhl-accenture-com i brought this down and built on my local and the full screen is still not showing as full screen, the video is still just on the left side. Additionally the growth is still being observed in the resource attachment area and on the homepage, like taking up the whole page. it is weird.

JJ-Muller (approved):

PR #366, opened 2021-04-20
T4T-712 populate the landing page from the API

jim-ruhl-accenture-com:
Also, the stories 713, 714 and 715 for the actions in the call to action section are included.


JJ-Muller (changes requested):
@jim-ruhl-accenture-com I am sorry I didn't catch this before, but what do you think about changing the variables that are in the form of "resourceType%" for landing pages instead to "landingType%" form in order to differentiate from landing pages within T4T and the resource pages within T4T.

webapp/src/app/landing/landing.component.html
```diff
@@ -27,26 +24,204 @@ <h1>
...
+        </ul>
+    </div>
 
-</aside>    
\ No newline at end of file
+    <!-- Sample sections -->
+    <div *ngIf="landingPage.sampleSections && landingPage.sampleSections.sampleResources.length > 0">
+        <h5>Sample Playlists</h5>
+
+        <p>{{landingPage.sampleSections.header}}</p>
```
JJ-Muller, line 197: @jim-ruhl-accenture-com landingPage.sampleSections.header is the value indicated for the header of this section so that it does not have to be hardcoded - see line195

JJ-Muller (changes requested):

webapp/src/app/landing/landing.component.html
jim-ruhl-accenture-com: please review this change

JJ-Muller: @jim-ruhl-accenture-com did you push the commit? I am not seeing the change

jim-ruhl-accenture-com: The AWS error and the heading fix are done

JJ-Muller (approved):

PR #368, opened 2021-04-27
Update teaser view for ASR to include Instructional Use (if exists)

ramacsh:


webapp/src/app/data/resource/resource.service.ts
```diff
@@ -255,6 +255,7 @@ export class ResourceService {
       properties,
       attachments: json.attachmentsExist ? teaserASContent.attachments : [],
       thingsToConsider: json.thingsToConsiderExist ? teaserASContent.thingsToConsider : null,
+      instructionalUse: json.instructionalUseExist ? teaserASContent.instructionalUse : null,
```
JJ-Muller, line 258: @ramacsh since this is a required field, like the stepByStep field in the FA content you don't need write it conditionally. Recommend writing like the stepByStep field shown in the FA teaser section

jim-ruhl-accenture-com (approved):

webapp/src/app/data/resource/resource.service.ts
ramacsh: @JJ-Muller Made the change as recommended. Thank you!

JJ-Muller (approved):

PR #403, opened 2021-07-22
Bugfix/t4 t 836 show print

kb-smarterbalanced:
Ticket: https://smarterbalanced.atlassian.net/browse/T4T-836

Modified functionality - in Chrome/other browsers behavior is as specified. In Safari it is slightly different. 

Safari workaround: 
we change the print icon on the landing page to a "download", they click on it and it opens the modal like normal, everything is the same except they do not see the blue "print" button.  Thus, the primary solution for printing in Safari is for the user to download the file and then print. The secondary solution is that they click the Safari browser print icon that hovers above pdfs as a default and print via that flow.


JJ-Muller (approved):

ramacsh:
@kb-smarterbalanced In my local testing with safari, I see the download icon instead of print on the landing page, and the modal opens. The modal has a lot of whitespace life before so only the top of page 1 is previewable, and I’m not seeing the print icon on hovering above pdf for the secondary solution mentioned in the PR. 
Screen Shot 2021-07-23 at 10 03 08 AM

JJ-Muller:
@ramacsh we will record the issue you are seeing as a separate bug

kb-smarterbalanced:
Code merged to develop - https://smarterbalanced.atlassian.net/browse/T4T-836

PR #410, opened 2021-07-30
Update metadata styles to align

kb-smarterbalanced:
Fixed alignments on the side - https://smarterbalanced.atlassian.net/browse/T4T-844


JJ-Muller (changes requested):
Hi @kb-smarterbalanced the label size is being displayed in16px instead of the same size as the value and the height spacing increased unexpectedly.  when the class selector is removed from label in the scss (line43) the label is displayed with the font size that is defined in the class elements. (I don't understand why that would make a difference though.)
Additionally, I realized that the size has been 14.1 and it should be 14.41, apologies for the late miss, would appreciate if you could add that change in as well.
image

kb-smarterbalanced:
OK please review again @JJ-Muller

JJ-Muller (approved):

PR #431, opened 2021-09-08
Social media tracking/domain verification

kb-smarterbalanced:
For issue: https://smarterbalanced.atlassian.net/browse/T4T-919

Note - I didn't put this in its own component because Angular best practice is to add <script> tags to index.html.


webapp/src/app/layout/footer/footer.component.html
```diff
@@ -60,3 +60,4 @@ <h3>Policy</h3>
     <span>© 2021 The Regents of the University of California</span>
   </div>
 </div>
+<div>Where is this?</div>
```
JJ-Muller, line 63: @kb-smarterbalanced I think this note is a stowaway in this PR :)

JJ-Muller (changes requested):

webapp/src/app/layout/footer/footer.component.html
kb-smarterbalanced: This is why we do pull requests :)

JJ-Muller (approved):

PR #438, opened 2021-09-15
update aria label

kb-smarterbalanced:
This is the request that was previously approved, but I had put in the wrong ticket (wrong ticket was T4T-869) I have now put in the right ticket! T4T-870 - https://smarterbalanced.atlassian.net/browse/T4T-870

Please review again and re-approve. Sorry for any confusion.


ramacsh (approved):
@kb-smarterbalanced Tested on local by confirming the attribute 'aria-label=main' shows up on all the home, search and landing pages

PR #447, opened 2021-09-24
Fix blue border on card

kb-smarterbalanced:
Jessica and Shankar - sorry for the late change, but we should get this in so we don't lose our styling! https://smarterbalanced.atlassian.net/browse/T4T-945


kb-smarterbalanced:
For ticket - https://smarterbalanced.atlassian.net/browse/T4T-945

JJ-Muller (approved):

PR #488, opened 2022-02-24
1.16.0 ammended back onto develop

JJ-Muller:


JJ-Muller:
no need this time. Develop is moving forward with node 14 and merging node 10 could cause compatibility issues

PR #508, opened 2022-04-04
Features/t4 t 1062 ga search

kb-smarterbalanced:
Adding numbers


kb-smarterbalanced:
This is for: https://smarterbalanced.atlassian.net/browse/T4T-1062 - just needed to add numbers to the regex so they show in keywords.

PR #520, opened 2022-04-30
Hotfix/t4 t 1092 broken downloads

kb-smarterbalanced:
Updating request


webapp/src/assets/config/config.dev.json
```diff
@@ -13,7 +13,7 @@
   "api2pdfAuthorization": "",
   "api2pdfIsDockerVersion": true,
   "interimItemPortalUrl": "https://interim-uat.smarterbalanced.org/BrowseItems",
-  "enableAnalytics": false,
```
JJ-Muller, line 16: @kb-smarterbalanced should this be put back to false in dev?

JJ-Muller: it will be used in relation to GA4

JJ-Muller (approved):

